PR #8, opened 2023-12-02
Setup Doxygen documentation

jkotrc:
Having an HTML documentation page for OpenStarbound can be a helpful tool for all contributors.  Doxygen automatically creates various documentation formats from specifically styled comments. This commit converts some existing dev comments to doxygen format so that they appear on the page.

To generate the page, install and run `doxygen` in the root of the project. It can then be accessed in `doc/html/index.html`

[See doxygen reference](https://www.doxygen.nl/manual/commands.html) for some supported commands. Sections can be made for specific issues and I recorded some basic observations in `\attention` blocks.


Novaenia:
Hi - ArmorWearer doesn't do any disk operations, the diskStore/diskLoad functions just do Json conversion. Those function names are kinda misleading though, so I can see why you thought that.

PR #12, opened 2024-01-03
Move instruments to their own audio group and volume setting

Bottinator22:
Some users might like this, either for muting instruments or for prioritising instruments over other sounds.

Todo: add instrument volume to options menu, currently without editing starbound.config this causes all instruments to just be always full volume

Note: I am a bit of a C++ noob. There may or may not be a better way of implementing this rather than patching all RenderCallbacks to include a function for adding instrument-related audio.


Novaenia:
gave it a slider and merged, thank you!

Bottinator22:
Informative

PR #14, opened 2024-01-16
Fixed OpenStarbound's memory ballooning bug

FezzedOne:
Checking for `m_data` on line 247 causes unnecessary `malloc`s on line 259, which causes a lot of memory fragmentation and thus the memory ballooning in OpenStarbound. This pull request fixes that. Note that `Image::reset` *doesn't* get called on an object that's moved *from*, but instead on the object moved *to*. `Image::~Image` is what gets called on moved-from objects.


FezzedOne:
Whoops. Managed to pull in unrelated commits from Bott by accident. Re-submitting.

PR #15, opened 2024-01-16
PR: Memory leak fix for OpenStarbound.

FezzedOne:
Checking for `m_data` on line 247 causes unnecessary (and unfreed) `malloc`s for empty image data on line 259, which causes a lot of memory fragmentation and thus the memory ballooning in OpenStarbound. This pull request fixes that. Note that `Image::reset` doesn't get called on an object that's moved *from*, but instead on the object moved *to*. `Image::~Image` is what gets called on moved-from objects.

### DIFF
--- a/source/core/StarImage.cpp
+++ b/source/core/StarImage.cpp
@@ -244,7 +244,7 @@ void Image::reset(unsigned width, unsigned height, Maybe<PixelFormat> pf) {
   if (!pf)
     pf = m_pixelFormat;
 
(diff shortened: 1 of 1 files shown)


FezzedOne:
Oh, and according to my tests, it does indeed reduce RAM usage upon unloading worlds, while the other fix that was merged fixes the rest. Both fixes together should solve the memory ballooning issue.

kblaschke:
There's also an actual memory leak in the Image class, see my fix in commit 42fc1d671

PR #17, opened 2024-02-13
seems to fix the memory issue

yzh5606:
by making some changes in StarLua.cpp
don't know if it will cause any bug
don't know if the /run cmd really fixed (cuz i don't know how to use the command)


lumikalt:
Thanks.

PR #20, opened 2024-02-19
A few small fixes

kblaschke:
One fixing an issue using memcpy to copy overlapping memory regions, the other fixes a Linux/macOS linker issue when building the tests in Release mode.


kblaschke:
Found and fixed a memory leak in `Image::readPngMetadata()`, which accounted for many megabytes of lost RAM over time.

PR #36, opened 2024-03-13
Add clipboard Lua bindings.

ErodeesFleurs:
Add clipboard Lua bindings, which is basically consistent with the clipboard Lua bindings of StarExtensions.


ErodeesFleurs:
Now the ClipboardLuaBindings are only available when "safeScripts" is false.

PR #43, opened 2024-03-19
Add the Lua function for sending packet.

ErodeesFleurs:
**void universe.sendPacket(ConnectionId clientId, String packetTypeName, Json args)**
Currently, only a few packet types are supported for sending.


ErodeesFleurs:
> I think it would be easier to work with if the packet arguments were a Json object rather than an array, but that can be added later anyways. Thank you!

I will improve it ^^

Issue #87, opened 2024-07-17
Controller analog stick drift (?)


Campoyo:
I noticed that the game recognizes the left stick of my controller, but there's a problem, if the controller has a small tilt to either left or right(Like any controller with a year of use), the character starts to move very slowly towards that direction making it impossible to crouch, sit or sleep, it happens even if I Alt+Tab to another window. Maybe the problem would be solved if we had a way to control the dead zone of the stick, or if we can disable that function.

I use a PS4 controller with DS4Windows to map the keyboard on the controller. Changing the dead zone in that software doesn't solve the problem, and testing with an Xbox 360 controller the problem still persists.

PR #94, opened 2024-08-03
Fix LeftStick movement StarClientApplication.cpp

RohanBhattacharyya:
It no longer has that drift. The drift is caused by the controller, and this simply turns up the threshold of where the game actually accepts the input as a movement.

This fixes the issue found in [this issue.](https://github.com/OpenStarbound/OpenStarbound/issues/87)


Campoyo (commented):
The change does indeed solves the problem.
but still I hope we could have an in game option to change the deadzone or a button to disable the analog movement completely in the future.

PR #101, opened 2024-08-22
Post-processing layer support for mods

Bottinator22:
Code could probably use some improvements in places, but it doesn't hurt my FPS or memory usage very much, so it's probably fine.

Here's some example mods for how this can be used:
Recolour: 
[recolour.zip](https://github.com/user-attachments/files/16703238/recolour.zip)
Bloom:
[bloom.zip](https://github.com/user-attachments/files/16703240/bloom.zip)

Incompatible with super-sampled AA, enabling that at all completely breaks rendering with these even after it is turned off. Todo: either disable/prevent enabling of SSAA while these are active, or fix SSAA for these.

### DIFF
--- a/assets/opensb/client.config.patch
+++ b/assets/opensb/client.config.patch
@@ -9,5 +9,6 @@
   "deployCinematicBase" : {
     "scissor" : false,
     "letterbox" : false
-  }
+  },
+  "postProcessLayers": []
 }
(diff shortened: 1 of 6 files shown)


Bottinator22:
Screenshot_20240821_193136
(screenshot with both of these on)
Screenshot_20240821_193705
(Moonquill sword with bloom)

PR #114, opened 2024-09-13
config option to disable scrolling action bar

SilverSokolova:
Setting `inventory.scrollActionBar` in starbound.config to false (defaults to true) prevents the player from scrolling the action bar with the mouse wheel. Required for a custom action bar I'm working on, since I need the default scrolling to be disabled


SilverSokolova:
Closing because this behavior is no longer needed as we can dismiss the action bar altogether to get the same result!!

PR #165, opened 2025-01-17
dungeon generation log messages include dungeon name

patmanf:
hi


patmanf:
what

DarianLStephens:
If a mod edits a dungeon like this, I'm not aware of any way of finding which one without searching every mod.
If it's possible at this point in the code, it would be nice to have it tell us what mod it comes from for the error messages.

PR #181, opened 2025-02-10
Switch to Ravi Lua

Bottinator22:
I've put Ravi behind another flag (STAR_USE_RAVI) which is currently only enabled on Linux.
Binaries for Mac/Windows will need to be provided by someone else. The binaries I used were built from
https://github.com/dibyendumajumdar/ravi
(after commit 5625033e935d3edbf104a4dea222ecb94dc6b997)
Upon building Ravi, take the libraries from `ravi/lib` and place them in `/lib/<os>`.

After building SB, you'll need to place Ravi's libraries next to the SB executable, similarly to Steam and Discord.
(for example, I place libravi.so, libravicomp.so, and libc2mir.so in the same folder as the Starbound executable)

So far I've found only one inaccuracy to 5.3 (which has been fixed by the aforementioned commit), but otherwise it seems to work fine in-game.
I've placed a simple configuration script at `/scripts/opensb/ravi_config.lua` that is currently set up for a constant configuration (currently, JIT on, auto mode off).


Bottinator22:
Most importantly, Ravi's API need to be reviewed and possibly culled for security reasons before this change is pushed into a release.

Bottinator22:
So far, `compiler.compile` seems to be the only one that should be disabled.

goodusername123:
If the implementation of Ravi goes well is it planned to pull in the rest of it and have it as the default for all builds (default as in base Ravi itself not saying the JIT)? because Ravi makes a bunch of changes to the core of Lua 5.3 and while it has the goal of being backwards compatible with pre-existing Lua 5.3 code it introduces features that are not compatible with base unmodified Lua 5.3 such as it's optional static typing system and the `defer` statement (although atleast `defer` can be patched into base Lua 5.3 https://github.com/dibyendumajumdar/ravi/blob/master/patches/defer_statement_for_Lua_5_3.patch) which would mean that any mods that start replying on Ravi specific features won't function in base Lua 5.3 which is a problem if not every build/OS is on Ravi.

Additionally Ravi also randomly introduces Mike Pall's [Lua BitOP module/library](https://bitop.luajit.org/) which isn't documented in it's manual and seems to have been just done to allow for the implementation of a third party dynamic code assembler which never ended up getting added to Ravi, so it might be worth it to disable this library as having a random extra bitwise library that only works on 32-bit ints and is prefixed with "bit." is kinda confusing in the context of Lua 5.3 which has native 64-bit integers and operators for doing bitwise ops on them (the ".bit" lib made more sense in older versions of Lua) so the introduction of this library should maybe or probably be done as a separate decision from Ravi. The relevant line to disable it is here: https://github.com/dibyendumajumdar/ravi/blob/5625033e935d3edbf104a4dea222ecb94dc6b997/src/linit.c#L56



Also something semi off-topic that ill ask in this comment since idk if this would be relevant for a full issue post but why is there a .cmake file in the cmake folder looking for [Lua 5.***2***](https://github.com/OpenStarbound/OpenStarbound/blob/main/cmake/FindLua52.cmake) (I'm a complete amateur when it comes to build systems so maybe this file isn't even used)

goodusername123:
Something else I have just noticed is that Ravi turns on `LUA_COMPAT_5_1` and `LUA_COMPAT_5_2` in "`luaconf.h`" (base Lua 5.3 has these off by default) and these should probably be turned off as they introduce a bunch of legacy Lua stuff that's only really relevant for projects that need some form of compact with Lua 5.1 and Lua 5.2. https://github.com/dibyendumajumdar/ravi/blob/5625033e935d3edbf104a4dea222ecb94dc6b997/include/luaconf.h#L305

Bottinator22:
The platform-specific Ravi is currently temporary as I don't have the means to set it up on Mac/Windows

Bottinator22:
Starbound also already requires adding Lua tables manually (I had to do so for ravi and compiler global tables) so it probably isn't using the bit table anyway

Bottinator22:
I've commented out the lines enabling 5.2/5.1 backwards compat.

Bottinator22:
As for the CMake file for 5.2, that's likely a leftover from an older version of Starbound before CF integrated Lua 5.3. Base Starbound doesn't use this either.

PR #189, opened 2025-02-21
[fix] Use bundled libraries first

leohearts:
fixes ./starbound: symbol lookup error: ./starbound: undefined symbol: SteamInternal_SteamAPI_Init

caused by inconsistent versions of `libsteam_api.so` .


WarpspeedSCP:
this doesn't seem to work.

PR #193, opened 2025-02-22
Make assets load only once

WereTech:
This will include UGC content if present and not require a reload that also loads non-UGC content again, causing certain cases like the post load scripts to run a second time. 

I have not noticed any issues, loading to main screen and into a character works fine with no warnings about missing content. I don't know if this has any potential issues somewhere. 

Additionally it skips the "Mods" state because of the changes I made, unsure if they were necessary.


WereTech:
Closing in favor of a better method.

PR #197, opened 2025-02-26
make scrollareas updating their children optional

patmanf:
it's kindve useful for scrollareas to be broken :3

mostly fixes stardust core scrolling


zetaPRIME:
right, I should probably still fix this on my end (that is, accept a PR and get around to pushing a release with the osb features)

PR #198, opened 2025-02-27
Ravi

Bottinator22:
Remaking this since I moved Ravi to a different branch of my PR fork.

I've put Ravi behind another flag (STAR_USE_RAVI) which is currently only enabled on Linux.
Binaries for Mac/Windows will need to be provided by someone else. The binaries I used were built from
https://github.com/dibyendumajumdar/ravi
(after commit 5625033e935d3edbf104a4dea222ecb94dc6b997)
Upon building Ravi, take the libraries from ravi/lib and place them in /lib/<os>.

After building SB, you'll need to place Ravi's libraries next to the SB executable, similarly to Steam and Discord.
(for example, I place libravi.so, libravicomp.so, and libc2mir.so in the same folder as the Starbound executable)

So far I've found only one inaccuracy to 5.3 (which has been fixed by the aforementioned commit), but otherwise it seems to work fine in-game.
I've placed a simple configuration script at /scripts/opensb/ravi_config.lua that is currently set up for a constant configuration (currently, JIT on, auto mode off).


WarpspeedSCP:
Nice, how noticable is the performance uplift?

scripts/linux/setup.sh
```diff
@@ -12,16 +12,18 @@ if [ -d /usr/lib/ccache ]; then
   export PATH=/usr/lib/ccache/:$PATH
 fi
 
+export VCPKG_ROOT=/home/bottinator22/vcpkg/vcpkg
```
WarpspeedSCP, line 15: you should probably remove this.

Bottinator22:
> Nice, how noticable is the performance uplift?

No noticable changes with the current state of this. However, this provides the necessary code to compile scripts in the future, making them much faster. We'll need to figure out how to do this compilation ahead of time, since Ravi's auto-JIT just causes constant lag spikes.

assets/opensb/scripts/opensb/ravi_config.lua
goodusername123, line 1: Would it make sense to also enable the generational garbage collector that Ravi has backported from Lua 5.4? as far as I understand it would be as simple as adding `collectgarbage("generational")` to "ravi_config.lua"

Emmaker:
This seems incredibly dangerous. Ravi compiles to machine native code, which is incredibly dangerous if not trusted and I can see someone finding an exploit that enables them to escape the Lua sandbox and damage the user's computer without needing safeScripts disabled.

Emmaker:
This potential threat is only exacerbated by the fact that it's a very small team working on this dialect of Lua.

Emmaker:
Hell, this doesn't even HAVE a sandbox. This can't be safe.

Bottinator22:
Due to how old this PR is, and the fact that it'd need significant reworks of several Lua systems to do anything useful, closing this.

PR #203, opened 2025-03-01
Bring new textbox callbacks

KrashV:


KrashV:
Sorry, not textboxes, not checkboxes :\

PR #206, opened 2025-03-02
Allow persistently preloading assets

Bottinator22:
Adds a file (`/preload.config`).
After assets are loaded, goes through this file and loads every asset listed in it, forcing those assets to be persistent.
Should allow reducing lag spikes in cases involving loading assets (such as, as already done here, inventory filters)

### DIFF
--- a/source/base/StarAssets.cpp
+++ b/source/base/StarAssets.cpp
@@ -24,6 +24,14 @@
 
 namespace Star {
 
+EnumMap<AssetType> const AssetTypeNames{
+    {AssetType::Json, "json"},
+    {AssetType::Image, "image"},
+    {AssetType::Audio, "audio"},
+    {AssetType::Font, "font"},
+    {AssetType::Bytes, "bytes"}
+};
+    
 static void validateBasePath(std::string_view const& basePath) {
   if (basePath.empty() || basePath[0] != '/')
     throw AssetException(strf("Path '{}' must be absolute", basePath));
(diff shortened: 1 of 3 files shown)


source/base/StarAssets.hpp
```diff
@@ -38,6 +38,15 @@ struct FramesSpecification {
...
 
+enum class AssetType {
+  Json,
+  Image,
+  Audio,
+  Font,
+  Bytes
+};
+extern EnumMap<AssetType> const AssetTypeNames;
+
```
bmdhacks, line 49: I don't think this is necessary.

bmdhacks (commented):
This looks very useful.

source/base/StarAssets.cpp
```diff
@@ -353,6 +361,20 @@ Assets::Assets(Settings settings, StringList assetSources) {
...
     m_workerThreads.append(Thread::invoke("Assets::workerMain", mem_fn(&Assets::workerMain), this));
+  
+  // preload.config contains an array of files which will be loaded and then told to persist
+  Json preload = json("/preload.config");
+  Logger::info("Preloading assets");
+  for (auto script : preload.iterateArray()) {
+    auto type = AssetTypeNames.getLeft(script.getString("type"));
+    auto path = script.getString("path");
+    auto components = AssetPath::split(path);
+    validatePath(components, type == AssetType::Json, type == AssetType::Image);
```
bmdhacks, line 372: I believe Images can also contain sub-paths, so this should be:

`validatePath(components, type == AssetType::Json || type == AssetType::Image, type == AssetType::Image);`

Bottinator22: Right, frames exist.

source/base/StarAssets.hpp
Bottinator22: I did this to allow loading *any* kind of asset dynamically. See StarAssets.cpp:369

I did this based on another instance I could find of dynamically accessing an Enum using a string. I'll see if removing that does anything, though.

bmdhacks: You don't need the extern declaration in the header file unless you're accessing AsseetTypeNames in other cpp files.  However a more C++ ish way to do it would be to move the AssetTypeNames declaration into the private: section of the Assets class and declare it static const that way it doesn't pollute the Star namespace.

bmdhacks (approved):
Looks good! This is gonna be very useful to my work too!

source/base/StarAssets.hpp
Bottinator22: I have committed this

PR #209, opened 2025-03-12
head rotation body masks

Bottinator22:
Screenshot_20250312_111922

(I also added a flag to ignore NaN/Infinity disabled warnings on Clang since they're really annoying when building)


Bottinator22:
(hold on, variable name moment)

PR #215, opened 2025-03-22
Block swapping

chililisoup:
Adds block swapping like in Terraria

https://github.com/user-attachments/assets/dd48bf06-3f13-472b-8410-5e875597573b

It adds a world client -> world server tile replacement packet to make it so attached objects don't pop off, but when connected to vanilla servers it just sends a regular tile damage packet.


Bottinator22:
if this can be done with Lua and have drops disabled, it'd definitely be useful on my TerraLib biome spreader (breaking objects is its main issue)

SilverSokolova:
Are you open to making this work if non-beamaxe items are equipped in the beamaxe slot? Some mods add scripted Lua manipulators and let you put them (and pickaxes/drills, or the vanilla energy pickaxe) in the beamaxe slot

PR #216, opened 2025-03-23
(WIP; semi-broken) block swap changes

SilverSokolova:
Since some mods allow the player to change their beamaxe slot item, I changed blockswap to work when a non-beamaxe item is equipped in the beamaxe slot, ie pickaxes (miningtool) and active items (vanilla energy pickaxe (this is what the primaryAbility checks are for), scripted activeitem manipulators). This check existed because of tool calls for the windup + firetime calculation between block strikes. I changed it to use the block's windup/cooldown time between strikes instead of the tool's, which is temporary until I or someone else finds a way to make it work for other item types

I also commented out the 'grab the strikeSounds of the player's tool and use it when blockswapping' because I couldn't get it to work with the energy pickaxe for now. I plan to fix these but will probably be busy for a while

### DIFF
--- a/source/game/items/StarMaterialItem.cpp
+++ b/source/game/items/StarMaterialItem.cpp
@@ -257,9 +257,6 @@ void MaterialItem::blockSwap(float radius, TileLayer layer) {
     return;
   
   Item* beamAxe = beamAxePtr.get();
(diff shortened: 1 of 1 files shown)


Bottinator22:
Why not make this accessible to Lua?

PR #223, opened 2025-04-01
Configurable post-process shader passes

Bottinator22:
(No, this is not an April Fools joke)
On the engine side, adds a Lua callback that allows changing the number of render passes for post-process layers based on a provided label. The means of keeping track of these labels is... not as clean as I'd hope, but pointers seemed to not work until reload for... some reason.
On the assets side, adds this functionality to loadconfig and the shaders menu. Options to change the number of passes are added as any other parameter, just set up differently. Note that a single post process group can have multiple layers with their own number of passes which may need their own sliders to configure this.

Updated bloom shader that uses this:
[bloom.pak.zip](https://github.com/user-attachments/files/19555004/bloom.pak.zip)

### DIFF
--- a/assets/opensb/interface/opensb/shaders/shaders.lua
+++ b/assets/opensb/interface/opensb/shaders/shaders.lua
@@ -184,6 +184,18 @@ local function addOption(data, i, added)
   end
 end
 
+local function updateParameter(parameter,oname,value)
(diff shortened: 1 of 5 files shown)


Bottinator22:
vcpkg brok

PR #228, opened 2025-04-23
remove oSB lua bind from vanilla docs

SilverSokolova:
player.say is already documented in the oSB docs. this removes it from the vanilla docs


Emmaker:
meow

PR #237, opened 2025-05-09
World actor entity callbacks

WasabiRaptor:
Makes a class intended to contain everything shared between the "actor" entities such as Players, Monsters, and NPCs primarily the use of the actor mcontroller and the status effect controller 

this is then used to add some more lua callbacks to world that can be used to read data on said entities that was always being networked, but would have required an entity message and handler if lua scripts wanted to read said data

having these callbacks will greatly simplify many scripts by no longer needing async messages and handlers for these

### DIFF
--- a/source/game/StarActorEntity.hpp
+++ b/source/game/StarActorEntity.hpp
@@ -1,0 +1,19 @@
+#pragma once
+
+#include "StarActorMovementController.hpp"
+#include "StarEntity.hpp"
+
+
+namespace Star {
+
+STAR_CLASS(StatusController);
+STAR_CLASS(ActorMovementController);
+
+// this is just used to have a base for what the game generally considers 'actors' as they all use the ActorMovementController, as well as have a StatusController
+// theres potentially more things shared that could be moved here
+class ActorEntity : public virtual Entity {
+public:
+  virtual ActorMovementController* movementController() = 0;
+  virtual StatusController* statusController() = 0;
+};
+}
--- a/source/game/StarMonster.cpp
+++ b/source/game/StarMonster.cpp
@@ -868,4 +868,13 @@ Vec2F Monster::questIndicatorPosition() const {
   return pos;
 }
 
+ActorMovementController* Monster::movementController() {
+  return m_movementController.get();
+}
+
+StatusController* Monster::statusController() {
+  return m_statusController.get();
+}
+
+
 }
--- a/source/game/StarMonster.hpp
+++ b/source/game/StarMonster.hpp
@@ -18,6 +18,7 @@
 #include "StarLuaComponents.hpp"
 #include "StarLuaAnimationComponent.hpp"
 #include "StarLuaActorMovementComponent.hpp"
+#include "StarActorEntity.hpp"
 
 namespace Star {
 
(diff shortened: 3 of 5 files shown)


Bottinator22:
You also missed a callback equivalent to `status.resource` to grab the proper value of a resource.

Not all resources have a maximum.

Bottinator22:
> the actor class is a good idea but the dozens of new world callbacks are what i do not agree with here - there are already too many "entitySomething" callbacks in the world table.

proposal: `world.entity<something>` table that returns a userdata table which holds the remaining callbacks specifically for that entity

Bottinator22:
world table as a whole kinda has a *lot* of things...
- more metadata callbacks (like `world.size`, `world.getProperty`)
- utility callbacks that even could've been done in Lua (`world.magnitude`, `world.distance`, `world.xwrap`, etc.)
- collision callbacks (`world.pointCollision`, so on)
- all those entity callbacks
- entity type-specific entity callbacks (like those for containers)
- spacial logging callbacks (`world.debugLine` and so)
- unique entity callbacks (`world.loadUniqueEntity`, `world.findUniqueEntity`)
- loading-related callbacks (`world.loadRegion`, `world.regionActive`)
- some screen-related callbacks (`world.isVisibleToPlayer`, `world.clientWindow`)
- structure-related callbacks (`world.objectAt`, `world.material`, `world.mod`, `world.liquidAt`, dungeon stuff)
- universe flags
- various other details (wind levels, undergroundness, ocean level, gravity)

WasabiRaptor:
my main argument against the "too many callbacks" is, why is that an issue, why is having more things accessible to script an issue because I think the question should be why should it *not* be there and frankly, I see no reason for them not to be there.

All of these values are already being networked, while I personally only use a few of these, I do see there is some application for it, else there wouldn't be callbacks for these in mcontroller or status, and I copied essentially all of them for the sake of not being inconsistent and not bringing up the question "we can fetch a stat from an entity why can't we fetch a resource?" and etc. This is somewhat already the case because we can read health with a callback already, but we can't read any other stats despite the entire thing being networked.

there have been too many cases where in script, it would be incredibly nice to avoid doing an entity message and callback to fetch a value from another entity, which introduces async behavior, which is especially bad for say, a projectile where we want something to happen on hit, which is conditional to say a stat or resource or property of the target

WasabiRaptor:
> > the actor class is a good idea but the dozens of new world callbacks are what i do not agree with here - there are already too many "entitySomething" callbacks in the world table.
> 
> proposal: `world.entityData` (or similar) callback that returns a userdata table which holds the remaining callbacks specifically for that entity or, a callback table as a property of `world`, so `world.entity.stat`

This I feel, might be the best one, however! I do not know how to do that

WasabiRaptor:
Relaying things I said in the discord after making some attempts to return a userdata table with callbacks, while it is likely possible, it is outside my current skill and understanding of how I would be able to do that, and I do not think it is worth the effort for a number of reasons

and while I could continue to try and implement such behavior, I do not think it is worth the added complexity and confusion and potential danger of having stale pointers even if they are weak pointers 

The current implementation of this feature already works, it is consistent with how the existing callbacks in world look and behave and therefore are easy to understand for how lua modders already use world functions once I write them in the docs

quoting Zygan as he puts it rather clearly
> I think that the idea of having a little userdata for each entity that has all of the callbacks, is what the api should ideally look like yes
> but it already doesn't look like that, and it's better to keep things consistent
> because like... what do you do with the existing world entity callbacks?
> 
> - move them into the new system? breaks compatibility with existing scripts
> - leave them where they are? now you need to do different things for different callbacks somewhat arbitrarily, that sucks
> - duplicate them?? now it's just confusing that you can do the same thing multiple ways...

Issue #258, opened 2025-06-03
Add "camera look" permanent option


aj8k:
H! First of all I want to say thank you for the work! Haven't played for 5-6 years and it was very nice to return back and see such a huge change in modding scene with sources being available and actually supported. You guys are amazing ❤

I really liked the work you have done with key binding for zooming. It helps a lot, a small change with huge impact. There's one more thing that could enhance players' experience even further: an option for the **"camera look" feature to be constantly turned on**.

This would greatly improve the quality of life for players who prefer a persistent camera view, allowing for smoother exploration and building without constant key presses.

For personal use, I've actually implemented a small, hardcoded solution to achieve the "camera look" constantly on, and that part was relatively easy. However, creating a proper configuration and integrating it into the settings, specifically trying to add UI to the "graphic" settings, appeared to be more difficult as the menu window was not responsive. Perhaps somebody with more experience in starbound UI development could do it?

Thanks for considering this!

aj8k:
I just tried to make `ctrl` button a toggle instead of press action. Will test it. Maybe it's a good alternative. No UI changes required in such case.

PR #259, opened 2025-06-03
Refine camera offset

aj8k:
Keeps player visible during “camera look” with cursor (pressed `ctrl` and moving mouse). Previous behaviour could totally hide player character, depending on zoom level and aspect ratio. This change should fix it.

It may be beneficial for cases when player constantly press on `ctrl`. Have played with it constantly turned on for several hours, works good so far.

Issue: https://github.com/OpenStarbound/OpenStarbound/issues/258

Before:

https://github.com/user-attachments/assets/ee412b0f-19f0-43ba-b5cf-ce2d6c40887f

After:

https://github.com/user-attachments/assets/5f53b865-cecd-4f02-831d-888dc49ac2c3

### DIFF
--- a/source/client/StarClientApplication.cpp
+++ b/source/client/StarClientApplication.cpp
@@ -1209,8 +1209,6 @@ void ClientApplication::updateCamera(float dt) {
 
   auto assets = m_root->assets();
 
-  const float triggerRadius = 100.0f;
-  const float deadzone = 0.1f;
   const float panFactor = 1.5f;
   float cameraSpeedFactor = 30.0f / m_root->configuration()->get("cameraSpeedFactor").toFloat();
   cameraSpeedFactor /= (dt * 60.f);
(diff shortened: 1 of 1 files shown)


Novaenia:
this should probably be mod-configurable rather than the default, as it can be considered a nerf to vertical look range

rexeightyseven:
don't see problems here, the camera let's you zoom out further on verical axis to let you see more up and down with 16x9 aspect ratio screen, this allows you to see as much as on the horizontal axis, changing this should be only optional.

the only problem with fact that you can hide your player character outside the screen happens in Frackin Universe for example, when you get Unbreachable Darkness debuff or Insanity both give vignette effect around your character and this effect is not following camera but player character so if you move camera high or low enough the effects disappear until you see your character, things like that could be fixed but camera zoom range is very helpful so I rather not see it changed.

PR #260, opened 2025-06-05
Item count display chainges

grbr404:
Improved the large stack amount display
Format:
  1 - 999
  1000 - 9999999 = 1K - 999K
  1000000 - 9999999999 = 1M - 999M
  1000000000 - 9999999999999 = 1B - 999B
  (up to Sextillion)
This allows us to display all possible stack amounts in at most 4 characters,
Making large stacks not look as bad.

Issue encountered:
Just now I saw that it only works on Linux build at least on my fork and I don't get why.

would be huge help if someone could fix the build issue if its not only in my fork as i dont know what to do to fix it.

### DIFF
--- a/source/windowing/StarItemSlotWidget.cpp
+++ b/source/windowing/StarItemSlotWidget.cpp
@@ -196,8 +196,8 @@ if (m_item->count() > 1 && m_showCount) {  // we don't need to tell people that 
         formattedCount = toString(m_item->count() / 1000000000) + "b";
     } else if (m_item->count() >= 1000000) { // Million (m)
         formattedCount = toString(m_item->count() / 1000000) + "m";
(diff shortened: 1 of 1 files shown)


Novaenia:
the two largest are impossible to attain as they are past the maximum u64 value of 18446744073709551615

grbr404:
ups gona fix that

Novaenia:
additionally i think it should start displaying **k**'s >= 10000, as the count still fits into 4 characters until that point

grbr404:
ok and i should remove the 4 characters check as otherwise 18446744073709551615 would get cut of

grbr404:
all builds now succesfuly build on my fork.

Novaenia:
took a look afterward and decided to redo it to handle decimals too - despite that effectively replacing the original code, you'll still be credited for it in the next release's notes

PR #263, opened 2025-06-07
Add (GNU/)Linux ARM port

### LINKED ISSUE #208: Make builds for ARM
With the rise of ARM, I think that official builds for this architecture should exist. Without them, users of ARM systems (I personally prefer Nintendo Switch because of its form factor).

lotigara:
Compiles on my Nintendo Switch running Ubuntu **GNU**/Linux 24.04 (BTW, does it compile on anything other than GNU?).

This PR is a draft because OpenStarbound itself prints that architecture is `unknown` in the pause menu (any suggestions?). Also because I didn't compile it with Clang and write a CI script.

It should fix #208.


grbr404:
It prints "unknown" as the CMakeLists.txt couldn't detect the architecture which it calls "star_system". You can find the code in line 73 and below it. It's located at source/CMakeLists.txt. I assume that the goal is for it to identify as Linux, as that would follow standard identification of architecture, which is used for example for Mac (Intel and ARM both identify as Mac). Making sure the  "star_system" is Linux is important for the compiler to decide whether to include Linux-specific code, which the arm version would need.

atom-skillz:
Tested the commit on top of the current origin/main. It compiles, builds and runs more or less well (for the PineTab2). But... it fails to load any mod and ignores the current universe and players (maybe it is related to the huge amount of "Mutex took too long" exceptions it's getting on the PT2).

lotigara:
> Tested the commit on top of the current origin/main. It compiles, builds and runs more or less well (for the PineTab2). But... it fails to load any mod and ignores the current universe and players (maybe it is related to the huge amount of "Mutex took too long" exceptions it's getting on the PT2).

Hmm, I played this commit with Frackin' Universe and it works fine...

atom-skillz:
Might be some unrelated error obtained by a race condition on the engine. Got something similar in https://github.com/OpenStarbound/OpenStarbound/issues/302 but was in x86.

SophiaSPRKLZ:
just compiled it on arch using termux with chroot after trying the portmaster version and realised how poor the performance was even at 480p. im shocked honestly, frackin universe and arcana run fine. even in busy areas at 1080p+. this really does need adding tbh

gammamonbiscuit:
Using this for my osb server with 130+ mods, works perfectly.

lotigara:
> It prints "unknown" as the CMakeLists.txt couldn't detect the architecture which it calls "star_system". You can find the code in line 73 and below it. It's located at source/CMakeLists.txt. I assume that the goal is for it to identify as Linux, as that would follow standard identification of architecture, which is used for example for Mac (Intel and ARM both identify as Mac). Making sure the "star_system" is Linux is important for the compiler to decide whether to include Linux-specific code, which the arm version would need.

After almost 1 year, I found the reason...

OpenStarbound's build system uses compiler's pre-defined preprocessor macros for architecture detection.

Look at this line:
https://github.com/OpenStarbound/OpenStarbound/blob/37d5ad16f07e9cad116d2391a157ad141e581e4a/cmake/TargetArch.cmake#L31
This code only searches for `__arm64__` macro despite it is not defined on Linux.
So the solution should be the following change:
```diff
-#elif defined(__arm64__)
+#elif defined(__arm64__) || defined(__aarch64__)
```

gammamonbiscuit:
Tested building with this change, it works

lotigara:
Ready to review!

lotigara:
@Novaenia

PR #268, opened 2025-06-14
Added a dedicated Missing Mod item 

grbr404:
This removes the possibility of players accidentally placing their items, effectively losing them.

I made it so it still uses `genericItemStorage`. I initially wanted to replace it with `missingModStorage`, but it seems better to use the old parameter for compatibility with the old system.


Bottinator22:
OpenStarbound needs to be vanilla compatible and able to join vanilla servers without issue. A new item type is not vanilla compatible and will cause missing item errors for other players.

PR #281, opened 2025-07-05
A unique but similar and recognizable Logo/Icon for OpenStarbound

grbr404:
While working on making OpenStarbound look better on Linux in one of my forks, I realized that we don't really have any logo that represents OpenStarbound. So, I made an icon/logo that still looks like a Starbound logo but has its own unique visual identity. I hope you like it.

If you think this icon is a good fit for the project, please feel free to merge it. Additionally, you're welcome to use the icon for the GitHub organization if you find it suitable.


grbr404:
Here are a few reasons for why I made an icon in the first place:

* I made this icon so this project has an icon that is not just the Starbound logo, as this could be very helpful if we ever want to build a flatpack as discussed in an issue someone made. (I have started to look into building a flatpack for OpenStarbound, but this is still going to take a while)

* As having a proper, kind of original icon could make distribution of a flatpack on places like Flathub a lot easier (Flathub is basically an app store for Linux, which makes installing and updating basically super easy).

* And having an icon that we have the SVG files for makes implementing the icon for all platforms a lot easier, as to my knowledge only Windows currently has an icon included with the app.

I would appreciate getting some kind of feedback on this icon suggestion. 

As always, thanks for the great work you do for the Starbound community.

PR #291, opened 2025-07-13
Added dithering to reduce color banding in lightmap.

grbr404:
It makes the lighting look a lot better as it makes the color banding a lot less.


Novaenia:
I think this could just be a regular mod that patches the shader with a load script. when i tried this, I noticed that the dithering is applied to areas of the light-map that don't have even banding, that being mostly completely dark and fully lit areas. in my testing, this negatively affects the file size of .png screenshots that I took. while it's possible to tweak the shader to skip dithering when the sampled light is nearly-black/white, that wouldn't help the cases where a world has an ambient light level set.

image

PR #295, opened 2025-07-17
Humanoid animations

WasabiRaptor:
Grants the humanoid species the ability to be configured to use the Networked Animator that the rest of the game uses for Monsters Objects and Vehicles, as well as granting the ability to have humanoid configs to be run through a build script on the initialization of a humanoid, so that values within could be modified based on identity values

HumanoidIdentity also now contains a JsonObject "extra" for arbitrary data, it is networked separately from retail's identity values, it is intended to carry any extra data for species that may end up using the networked animator features

Several behaviors inherent to humanoids are still handled by the engine for the sake of not causing more network traffic than necessary by the animator, as well as from contexts that would not have a script available, as the animator now also has some functions for triggering animations or setting tags locally on the client, however the animator callbacks are still added to the master entity's script contexts, as well as having a client animation script. However, while such things are hardcoded, what a creator configures their animation to do with these is entirely up to them.

I will try to write proper documentation later
but a general rundown is, a number of offsets and rotation centers can be configured to a part point in the animator
what parts the hand items are parented to can be changed
list of arbitrary statetypes to set based on humanoid state
list of arbitrary statetypes to set based on humanoid state, when moving backwards
list of arbitrary statetypes to set based on humanoid emote
list of arbitrary statetypes to set based on portrait mode

head rotation arm rotation and backCosmetic transformation groups are hardcoded 
statetype for held item being inside or outside hand is hardcoded 
statetype for arm being in idle or rotation sate is hardcoded, it will attempt to use the state for the given arm frame if available but default to a default if it isn't
it does effectively the same for dance frames, however if the requested dance has a state in the "dance" statetype it will also trigger that

equipped cosmetics are have animation tags they set based on their usual frameset and directives values, however they can define a table of humanoidAnimationTags to set any arbitrary tags in the animation, the keys of that table will have <slot> replaced with their relevant cosmetic slot number

while nothing is currently set to use this system there is a build script and a config included that was used for testing purposes
if a species has `"buildScripts" : ["/humanoid/opensb/build.lua"]` defined in its species config, it will use the build script to populate the animation config it outputs with the relevant animation values that should perfectly imitate how the base humanoidConfig could change things

NPCs now also support updating their identity with lua callbacks the same way players can


WasabiRaptor:
tried to changed netHumanoid to inherit from netElementSyncGroup rather than netElement, but the compiler didn't like that, so I just put a netElementSyncGroup inside it

PR #301, opened 2025-07-21
Lua core: backport Jump Table from Lua 5.4

goodusername123:
Backports the Jump Table implementation from Lua 5.4 which should provide around a 5% or so performance increase/boost to the core Lua interpreter for compatible compilers such as GCC and Clang.

**IMPORTANT NOTES:**
I don't actually have the correct setup to compile and test this myself sadly :( but I don't see why this wouldn't work.
Windows compiles won't benefit from this unless they are switched off of MSVC to Clang or something.

### DIFF
--- a/source/extern/lua/ljumptab.h
+++ b/source/extern/lua/ljumptab.h
@@ -1,0 +1,91 @@
+/*
+** $Id: ljumptab.h $
+** Jump Table for the Lua interpreter
+** See Copyright Notice in lua.h
+*/
+
+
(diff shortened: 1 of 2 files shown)


goodusername123:
There's a post online that goes over the concept and benefits of this (not specifically *for* PUC Lua's implementation but it's still the exact same core idea): https://eli.thegreenplace.net/2012/07/12/computed-goto-for-efficient-dispatch-tables

PR #304, opened 2025-07-24
player UI message callback

WasabiRaptor:
add a callback to let scripts be able to send arbitrary messages here


WasabiRaptor:
didn't know about the interface bindings

PR #311, opened 2025-07-28
Make Mcontroller callbacks be inherited for actor mcontroller callbacks

WasabiRaptor:
removes the need for duplicate callback definitions of base mcontroller callbacks so that the only new callbacks that need to be defined for the actor mcontroller are the actual new callbacks

(and the callbacks that need to set an additional value)


patmanf:
are you aware that actor mcontrollers now inherit the `accelerate`, `applyParameters`, `approachVelocity`, `approachVelocityAlongAngle`, `approachXVelcocity`, `approachYVelocity`, `collisionArea`, `collisionBoundBox`, `force`, `localBoundBox`, `parameters`, `resetParameters`, and `rotate` callbacks

vanilla didn't give actors those callbacks and i had assumed it was for a reason. but i did test them a little bit and they all seemed to work perfectly fine lol

PR #325, opened 2025-08-04
Fixes

WasabiRaptor:
let netHumanoid exist again because I'm pretty sure it is fine and the other person was just having issues from older versions

make sure that the element counts clear when net elements are cleared and make it consistent in how it counts them with the rules checked, don't know why I didn't just do that the first time

fix when it stores the humanoid parameters and npc description into overrides, as I didn't realize overrides wasn't always just a jsonObject and npcs that weren't initialized with overrides won't have it


WasabiRaptor:
found some misc minor things that had happened when re organizing some functions that I didn't end up seeing because I had removed the patch enabling the animator mode that was for testing

PR #326, opened 2025-08-06
Error handling scripts

WasabiRaptor:
Makes the resulting item from an error be determined by script, this allows a mod to say, delete items that have been removed from the game instead of there just being an eternal generic item, or convert the removed item into something else.

Also adds an error handling script for every entity type, objects that have been removed will now convert themselves into a generic item using their error handling script.

plenty of other practical applications for the ones for other entities such as converting npcs with species or npc variants that were removed into a different one (though the default script here doesn't do that)

same applies to players, monsters, vehicles getting their own script, these scripts are defined in a mod's metadata


Novaenia:
I think it'd be more consistent if rebuild scripts were defined in a /rebuilds.config file, rather than in mod metadata
(also the script file names don't need to end in Rebuild, the parent folder name already says enough)

PR #345, opened 2025-08-26
world.entity bindings

Bottinator22:
I have yet to test most of these bindings in-game, but they SHOULD work. A few of them might try to access the world and throw when used after the target entity has already been removed from the world, but NONE of them should segfault.

Adds a world callback, `world.entity(EntityId)`. This returns a userdata offering access to that entity. This userdata persists, and updates with the entity, so getting it more than once per entity is unnecessary. The entity is also held in memory until the userdata is destroyed. This does not cause any new memory leaks as the script remains responsible for cleaning up after itself.

Since the actual entity instance is linked to the userdata, `world.entity` run on the main player of a client is the only instance where an entity stored in this userdata can start existing again.

This system could be expanded on in the future to allow more direct control of entities, such as removing them or re-adding them to worlds via additional world callbacks.

For the entity bindings, I unified the two entity parameter callbacks into a single method, entity:getParameter, which also allows reading parameters from stagehands and projectiles for now. Should probably add monsters and vehicles to it later.

Though unrelated to what I'm doing here, this PR also enables properly overriding the scripts on projectiles, mostly to make my life easier.

I'll add docs after this is merged.

### DIFF
--- a/source/game/scripting/StarLuaEntityUserdata.cpp
+++ b/source/game/scripting/StarLuaEntityUserdata.cpp
@@ -222,62 +222,62 @@ LuaMethods<EntityPtr> LuaUserDataMethods<EntityPtr>::make() {
     // actor entity methods
 
     // status controller methods, they're networked anyway so might as well make them available to read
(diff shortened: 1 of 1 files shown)


Bottinator22:
hold on I should... probably rename those bindings to be more consistent with the rest

PR #356, opened 2025-09-02
Fractional interface scaling

Sharp-Edged:
Changed `interfaceScale` from an `int` to a `float`. 
Also partially changed `StarVector`, `StarRect` and `StarPoly` for more sensical behaviour when scaling integer typed containers by floats. 
Hopefully nothing is broken ¯\\\_(ツ)\_/¯

### DIFF
--- a/source/windowing/StarGuiContext.cpp
+++ b/source/windowing/StarGuiContext.cpp
@@ -4,7 +4,6 @@
 #include "StarMixer.hpp"
 #include "StarAssets.hpp"
 #include "StarImageMetadataDatabase.hpp"
-#include <iostream>
 
 namespace Star {
 


Novaenia:
fails to compile on Clang currently - keep in mind that scaling pixel art fractionally will appear choppy and uneven
interfaceScaleList needs to be restricted further - 0.5x increments would be enough
image

Sharp-Edged:
Yeah I think fractionally scaling pixel art is impossible without it looking janky unless we implement some custom pixel art scaling algorithm (though probably won't 100% fix the problem + not worth the effort). So I think we should keep the current increments and maybe eventually just add some "warning" which says that it will look ass if it's not an integer increment. Though the pixel art is usually small enough that it's not that noticeable if I'm not paying attention.

I think the main repo doesn't build on clang, it says the compile error is from `game/StarWorldClient.cpp` which I didn't modify.

Novaenia:
> I think the main repo doesn't build on clang, it says the compile error is from game/StarWorldClient.cpp which I didn't modify.

oh, i didn't see. i'll go fix that

PR #378, opened 2025-09-26
randomly assign wire colors for objects which don't have them defined

WasabiRaptor:


Novaenia:
this should be a mod instead

PR #391, opened 2025-10-27
Fix Drawable lua converter to be consistent

WasabiRaptor:
The drawable lua converter was inconsistent with the drawable constructor from json, leading to transformations being compounded with being centered as well as scaled when these values should do nothing when a transformation was already defined leading to strange offsets when using things like canvas:drawDrawables with parts that had transformations applied

also a commit fixing the chest cosmetic applying a transformation group in the wrong animations that I forgot to PR from a while ago


Bottinator22:
I recall making drawable code that used this. Although I never released
that code. This might remove backwards compatibility with some local
animation scripts.

Bottinator22:
Turns out this PR breaks vanilla.

Issue #292, opened 2025-07-13
[BUG] Version 0.1.14 cannot click buttons


RohanBhattacharyya:
Hello, in this version: https://github.com/OpenStarbound/OpenStarbound/releases/tag/v0.1.14, button clicks do not work anymore. For example, when I boot up OpenStarbound, hovering over buttons (such as singleplayer), does not show the hover effect, and clicking them has no effect. This is not an issue for previous versions.

For reference, I use an NVIDIA GPU on Arch Linux using Wayland.

Please let me know if there is any other information you need!

Novaenia:
please attach starbound.log. are you using any mods that alter the title screen?

RohanBhattacharyya:
[starbound.log](https://github.com/user-attachments/files/21207316/starbound.log)

RohanBhattacharyya:
No, I did not have any mods installed.

B4LiN7:
I think this is a (system) scaling issue. I'm on Fedora 42 (KDE Plasma) and have scaling set to 125% since my monitor resolution is 3440x1440. If I set it back to 100%, the game works fine.
Also, if you disable the hardware cursor, you may notice an offset between the in-game cursor and your actual cursor.
Using fullscreen mode also solves the issue.

RohanBhattacharyya:
> I think this is a (system) scaling issue. I'm on Fedora 42 (KDE Plasma) and have scaling set to 125% since my monitor resolution is 3440x1440. If I set it back to 100%, the game works fine. Also, if you disable the hardware cursor, you may notice an offset between the in-game cursor and your actual cursor. Using fullscreen mode also solves the issue.

Interesting! Why is it that this only applies to version 0.1.14?

Novaenia:
hi, is this still occurring in the new nightly build?

B4LiN7:
Sadly, it didn't fix the issue for me. Just for reference, I upload a video demonstrating what I see.
Although I didn't show it in the video, if you increase or decrease the scaling, the offset changes accordingly. If you go below 100%, the game cursor can move outside the window.
https://github.com/user-attachments/assets/19383dc2-67ba-4972-9f02-bda3ff1c935c

RohanBhattacharyya:
Can confirm the nightly still does not work. Here is the log.

[starbound.log](https://github.com/user-attachments/files/21236153/starbound.log)

RohanBhattacharyya:
[...]

PR #414, opened 2025-11-30
SDL_WINDOW_HIGH_PIXEL_DENSITY with scaling of mouse position

grbr404:
This re-adds SDL_WINDOW_HIGH_PIXEL_DENSITY and fixes issue #292 by calculating the scaled mouse position and using it if we are on Linux and are using Wayland. I have made this Wayland-exclusive, as to my knowledge, this issue only occurred on Linux environments that are using scaling on Wayland.


grbr404:
I'll keep this as a draft until I've looked into `setCursorPosition` and determine if scaling handling is also required.

PR #419, opened 2025-12-07
Added setCameraFocusEntity

KrashV:


patmanf:
add it to /doc/lua/stubs/openstarbound/player.lua too i think

PR #425, opened 2025-12-16
Proper UI Scaling for Cursor relative to the UI

danellos:
The mouse cursor previously stayed at a fixed size regardless of interface scale, causing it to appear too small on high-DPI displays when the UI is scaled up. This was due to cursor configs with "scale": 1 overriding the interface scale entirely.

This PR makes the cursor scale with the interface scale setting at 50% of the rate. The intention here is for standard display scales where you'd use lower UI scaling, the cursor remains at its native size, whilst on higher scales, it grows proportionally while avoiding becoming oversized. This follows a similar philosophy to how desktop operating systems handle cursor scaling at high DPI.

### DIFF
--- a/source/frontend/StarInterfaceCursor.cpp
+++ b/source/frontend/StarInterfaceCursor.cpp
@@ -56,9 +56,9 @@ Vec2I InterfaceCursor::offset() const {
 }
 
 float InterfaceCursor::scale(float interfaceScale) const {
-  // AUTO integer scaling: derive from interface scale
-  int intScale = static_cast<int>(interfaceScale);
-  return static_cast<float>(std::max(1, intScale - 1));
+  int baseScale = m_scale ? static_cast<int>(m_scale) : 1;
+  int intScale = std::max(1, static_cast<int>((interfaceScale + 1) / 2));
+  return static_cast<float>(baseScale * intScale);
 }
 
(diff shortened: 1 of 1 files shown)


Novaenia:
causes mixels at scales 2.0 and 4.0 and does not function with hardware cursor enabled, this should be its own setting
image

tphilipgh:
@Novaenia I see what you're saying--I managed to fix the inconsistent scaling for both hardware and software cursor and removed the weird mixels using integer scaling, but am a bit unsure of where to put the scale option in the graphics options menu. I wanted to place it where the current "Shaders" button is located, but I am not sure where the right place to move the Shaders button is and feel I would need your opinion there. 
image

[EDIT] This is an idea I have, but the background bars are a bit out of whack:
image

tphilipgh:
@Novaenia I kind of gave up fighting with the UI of the Graphics Options menu due to the background pngs not playing nicely. For now, I changed the mouse scaling from percent scaling (float) to simple integer scaling based on UI scale. This at least gets rid of the mixels you correctly noticed.

At lower UI scales (less than 3) the mouse cursor should remain the same, but as soon as it gets to 3x scaling and beyond, the mouse cursor will start scaling automatically using integer scaling.

Agreed that ideally later this should be an interface option.

PR #426, opened 2025-12-16
Add config var to skip checking for UGC updates

J0w03L:
Since the recent asset loading rework, my game has begun trying to check for updates seemingly indefinitely (i.e; steam does not show any downloads in progress, and the update check never finishes) despite having a relatively small 270 mods installed (at least compared to some people I know).

I'm not really sure why that's happening; I'm compiling from source and running on a Linux host, and don't mind helping to debug this, but for now, here's a PR that adds a setting for bypassing the update check.

I figured I'd throw this in here, not just as a workaround, but also for people who might just not want to have their game update everything just yet (for example, someone playing on the go with a limited mobile data connection).

Simply add `"skipUGCUpdates" : true` to either your `starbound.config` or `sbinit.config` (of course, with a comma at the end if not the final config in the file), and the game will not check for updates to UGC content.

### DIFF
--- a/source/client/StarClientApplication.cpp
+++ b/source/client/StarClientApplication.cpp
@@ -855,16 +855,17 @@ void ClientApplication::updateMods(float dt) {
   auto ugcService = appController()->userGeneratedContentService();
   auto configuration = m_root->configuration();
   bool includeUGC = configuration->get("includeUGC", m_root->settings().includeUGC).toBool();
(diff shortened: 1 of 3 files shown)


J0w03L:
#427 fixes the bug I was having here

Bottinator22:
Is this PR still needed then?

J0w03L:
Nah not really. I left it up as a possible idea (it would still be nice for people who wanna play the game on the go and have to super throttle their download speed because they're on mobile data and steam forces their workshop stuff to update............ but honestly kind of an edge case), it'll still exist if it's closed.

Issue #439, opened 2026-01-04
Main menu unresponsive on macos arm64 with high dpi display


WasabiRaptor:
Likely something do do with the high dpi display inherent to macbooks, the menu is rendering far too small, and the buttons are unresponsive when hovered over, however the mouse cursor is still being replaced and rendered at the correct position.

<img width="1440" height="900" alt="Image" src="https://github.com/user-attachments/assets/ae90f436-f069-477a-83d7-f3dfb0a472e8" />

grbr404:
So this seems like a repeat of issue #292. The mouse cursor is likely at the correct position because of hardware cursor is on but the game internally only uses the software one for input that's why nothing works. This is as you said because of high DPI I was not aware about the same issues being on Mac. The 2 options would be to either revert the change (#438) or to implement proper system scaling handling.

grbr404:
@WasabiRaptor I've now changed my fork where I implemented the proper conversion for mouse coordinates to also work on Mac. Could you test it? https://github.com/grbr404/OpenStarbound/actions/runs/20717659693 It should fix the mouse-related issues.

smartintruesdell:
The build from @grbr404's fork seems to work for me on my Macbook Pro M3 Pro 14inch.

grbr404:
@smartintruesdell first of all, thanks for the information, as I can't test if the Mac fix works. I also made an improved version that scales the UI. Its build is available at https://github.com/grbr404/OpenStarbound/actions/runs/21094987837 It would be great if I could get feedback on that, as I'm going to open a pull request for it.

smartintruesdell:
> It would be great if I could get feedback on that

It looks good on first launch. With the previous build I needed to go into Settings and up the interface scale to 3.5x or so to get it "usable" for my old man eyes. Now it's great out-of-the-box.

Previous build:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/18e9d9d8-792b-4898-84f1-b7c1ec84d02c" />

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/c31cbd49-81d1-416c-9082-ca691ee02d16" />

This build:

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/2ac3714f-d6b7-4121-a80c-2b664bc6df68" />

<img width="500" alt="Image" src="https://github.com/user-attachments/assets/d0364a8f-e6eb-4e6b-8d9e-0a67744b0ac7" />

[...]

PR #446, opened 2026-01-17
Sdl proper high dpi

grbr404:
This resolves issue #439, which is basically the same as #292, with the only difference being that it occurs on Mac and mentions the UI being too small.

This fixes both issues by converting mouse coordinates by multiplying them by m_displayScaleMouse, which is set to 1 by default. If we don't set it (on Windows and Linux X11), we get normal coordinates. If we do set it, we get logical coordinates (for Mac and Linux Wayland).

We get the display scale for all platforms so we can use it for interface scale. Interface scale is only scaled by the nearest full integer and is at least 1.

The only thing that should also be scaled like interface scale is the software cursor (not the hardware cursor, because it already gets scaled by the system). This is not implemented because the cursor does not use interface scale, and I think that needs to be addressed by someone first.

This has been tested on:
* Linux Wayland
* Linux X11
* Mac (tested by @smartintruesdell)

To see some great before-and-after pictures, you can look at:
https://github.com/OpenStarbound/OpenStarbound/issues/439#issuecomment-3764175317

### DIFF
--- a/doc/lua/openstarbound/root.md
+++ b/doc/lua/openstarbound/root.md
@@ -120,9 +120,11 @@ Sets a configuration value in `/storage/starbound.config` by path.
 
 ---
 
-#### `JsonArray` root.allRecipes()
+#### `JsonArray` root.allRecipes(`Maybe<StringSet>` filter)
 
 Returns all recipes.
+
+Optionally apply a filter for recipe groups.
 
 ---
 
--- a/source/core/StarFile_unix.cpp
+++ b/source/core/StarFile_unix.cpp
@@ -212,7 +212,7 @@ void* File::fopen(char const* filename, IOMode mode) {
 
 void File::fseek(void* f, StreamOffset offset, IOSeek seekMode) {
   auto fd = fdFromHandle(f);
-  int retCode;
+  StreamOffset retCode;
   if (seekMode == IOSeek::Relative)
     retCode = lseek(fd, offset, SEEK_CUR);
   else if (seekMode == IOSeek::Absolute)
--- a/source/game/scripting/StarRootLuaBindings.cpp
+++ b/source/game/scripting/StarRootLuaBindings.cpp
@@ -48,7 +48,7 @@ LuaCallbacks LuaBindings::makeRootCallbacks() {
   callbacks.registerCallbackWithSignature<float, String>("projectileGravityMultiplier", bind(RootCallbacks::projectileGravityMultiplier, root, _1));
   callbacks.registerCallbackWithSignature<Json, String>("projectileConfig", bind(RootCallbacks::projectileConfig, root, _1));
   callbacks.registerCallbackWithSignature<JsonArray, String>("recipesForItem", bind(RootCallbacks::recipesForItem, root, _1));
-  callbacks.registerCallbackWithSignature<JsonArray>("allRecipes", bind(RootCallbacks::allRecipes, root));
+  callbacks.registerCallbackWithSignature<JsonArray, Maybe<StringSet>>("allRecipes", bind(RootCallbacks::allRecipes, root, _1));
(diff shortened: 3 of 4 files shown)


smartintruesdell:
@grbr404 There's some weirdness on Mac:
When going to fullscreen, mouse events still work on the menu but the UI scale goes back to 1.0 and everything gets all tiny again.
When using a weapon with a targeting reticle like a gun, the targeting reticule is HUGE. Can't screenshot this because going to take a SS hides the cursor but it's the same height as the player character.

grbr404:
@smartintruesdell so what I assume is happening here is that maybe Mac does not apply scaling if the application is fullscreen. Can you tell me if the cursor (targeting reticle) is the same size in fullscreen as when not in fullscreen, because these issues seem to be video driver specific.

smartintruesdell:
@grbr404 When in fullscreen, it looks like it goes back to being pixel-perfect, normal player-relative scale.

```
[22:23:42.031] [Info] OpenStarbound Client v0.1.14.1 for v1.4.4 (macos arm64) Source ID: 5538631cc584831c74e20d8e525b297a204c2ab9 Protocol: 747
[22:23:42.031] [Info] Application: Initializing SDL Video
[22:23:42.145] [Info] Application: using Video Driver 'cocoa'
[22:23:42.145] [Info] Application: Initializing SDL Controller
[22:23:42.163] [Info] Application: Initializing SDL Audio
[22:23:42.191] [Info] Application: using Audio Driver 'coreaudio'
[22:23:42.210] [Info] Initialized Steam platform services
[22:23:42.210] [Info] Application: Creating SDL window
[22:23:42.233] [Info] Application: Creating SDL OpenGL context
[22:23:42.250] [Info] Application: Enabling adaptive VSync
[22:23:42.250] [Info] Application: Opening audio device
[22:23:42.302] [Info] Application: Opened default audio device with 44.1khz / 16 bit stereo audio
[22:23:42.319] [Info] OpenGL version: '4.1 Metal - 90.5' vendor: 'Apple' renderer: 'Apple M3 Pro' shader: '4.10'
```

grbr404:
@smartintruesdell I just found out that on Wayland, I experience the opposite issue: the input scales too much when fullscreen and UI scaling is normal. I don't know why this issue occurs on Mac either, so this is likely going to take a lot longer to fix. I'm going to make this a draft for now.

grbr404:
@smartintruesdell It would be great if you could test the newest version. It's at https://github.com/grbr404/OpenStarbound/actions/runs/21356657826

grbr404:
The new version no longer uses SDL_GetWindowDisplayScale to get the scale. It now uses SDL_GetDisplayContentScale(SDL_GetDisplayForWindow(m_sdlWindow)). This makes it so we get the scale of the display that we are on. This should fix UI scale reverting on fullscreen on Mac, as that was due to Mac not applying scaling to the window if it's fullscreen. We set mouse scaling now to 1 when in fullscreen to prevent the issue (over-scaling of mouse position) encountered on Wayland in the previous version. This issue only needs to be accounted for on Mac, as Wayland in this version uses the hint again that sets it to use physical coordinates.

grbr404:
@Novaenia Sorry for the ping but I have a few questions regarding this pull request.

First, what do you think is the best course of action regarding the HiDPI implementation? I believe this should now fully address all the issues, but I'd like to confirm it works across all systems if possible to avoid future issues. However, I can only verify that the Linux version works (except for X11, which has an unrelated fullscreen issue that I'll likely fix in the future).

Also, would you be able to test this on Windows if possible?

Below are two images showing the system scale being correctly applied to the game's UI:

scale100

scale175

Novaenia:
seems to work fine on Windows, i can merge it if you want

grbr404:
I think you can merge it.

WasabiRaptor:
Issue still occurs on mac

PR #451, opened 2026-01-22
Adjust materialConfig and modConfig in root callbacks to accept ID number or name

WereTech:


Novaenia:
needs to throw an error instead of returning nothing when passed a value type that is neither string or number
the Lua wrappers support Variant<>

WereTech:
Fixed to your suggested changes

PR #453, opened 2026-01-26
sub versioning data is serialized afterwards separately from the base versionedJson

WasabiRaptor:
the sub versioning worked perfectly fine for the lua callbacks where it could just be put into the container json object 

however, for the versioned Json that is serialized to binary previously I had simply smuggled it into the content when saving, and then removed it from the content when reading, this was fine for OSB, but when vanilla read back this it was not removing the versioning data from the content, and in the specific case of the tempworlds index, it was iterating over the content object and then reading the versioning data as if it was a tempworld and then failing.

New solution is, serializing the data properly outside the content in a manner that vanilla will simply ignore.

for all files that are just read and written to directly, it simply writes the versioning data to the end of the stream, when vanilla reads it back it simply reads the normal versioned Json data, and then won't care about the extra bytes following it.

The few cases it was used within the btree database for world sectors was only slightly more complicated, for the world metadata it simply appends it onto the end where it normally is serialized, and vanilla discards it. The complication was the entity sector, which serialized the entity versioned json as a list, so we simply serialize that list first, and then iterate over the list and serialize the sub versioning data

in all cases, each entry of sub versioning data is preceded by a version number so if it ever needs to change, that can be accounted for

I did personally test loading the world and players back and forth from OSB and retail repeatedly, and found no further issues.


Novaenia:
it could be future-proof to write a second json object at the end, with the subversion container inside that?
also, it still shouldn't be written at all when subversioning is not in use at all for the VersionedJson's specific identifier

WasabiRaptor:
yeah I'll switch it to be a jsonObject rather than serializing the subVersions directly

and I'll remove the currently unused osb identifiers

Issue #495, opened 2026-04-08
Running OpenStarbound on Haiku


Peppersawce:
I am trying to run OpenStarbound on [Haiku](https://www.haiku-os.org/), I am building it using https://github.com/OpenStarbound/OpenStarbound/commit/233344e7f98e138a83742621e613fe2dce922748 with some rough patches on top.

The binaries all compile, I put `sbinit.config` from the Linux scripts into `dist/` and copied `packed.pak` from my Linux/Steam copy of Starbound in `assets/`.

The `starbound` executable fails with an AssetException. `asset_packer` is also experiencing issues, although `asset_unpacker` works fine, I haven't checked the rest of the binaries.

Am I missing something? Like, installation steps or libraries?

Here's the log: [starbound.log](https://github.com/user-attachments/files/26577397/starbound.log)

Bottinator22:
You're missing oSB assets.
You seem to have added them incorrectly (I theorize either it's a file the game can't recognize or it's a broken symlink)
```
[19:59:09.205] [Warn] Root: Unrecognized file in asset directory 'opensb', skipping
```

You need it present either as a folder or as a .pak file. Since you're building directly from source, you can find that folder in `OpenStarbound/assets`

Peppersawce:
`Unrecognized file in asset directory 'opensb', skipping`
As you can see from the error, the opensb directory is already present in the `assets` directory and the binary finds it, but is being skipped due to an `Unrecognized file`.

Bottinator22:
It's likely either a broken symlink or it's missing the `.pak` extension (the game does check specifically for the extension). I don't recommend using the release `.pak` when building from source.
Or it was cloned incorrectly and is a file rather than a directory.

Bottinator22:
...considering you're running the game directly from the cloned source, the issue is likely that the `opensb` asset folder was somehow cloned incorrectly.

Send a screenshot from your file manager or similar of the contents of `OpenStarbound-main/assets`?

Peppersawce:
<img width="508" height="752" alt="Image" src="https://github.com/user-attachments/assets/8aa68fa2-c549-47fb-b461-372f2de2931d" />

`/assets` only has this folder and `packed.pak`

In the meantime I've been trying with 0.1.14's opensb.pak. This is the resulting log.

[starbound.log](https://github.com/user-attachments/files/26577777/starbound.log)

[...]

PR #496, opened 2026-04-09
Haiku port

Peppersawce:
Some patches based on my experience with #495, these changes allow OpenStarbound to build and run under Haiku 64bit, provided one has the correct libraries.
The main binary still crashes as described in the issue above but I think these initial bits of the port are good enough to be looked at.

`!S_ISREG(stbuf.st_mode)` in `StarFile_unix.cpp` in the directories check is probably not right but I didn't know what to put there tbh.

I think the changes in `CMakeLists.txt` are fine but I haven't checked on other platforms.

This is not the full amount of changes I'm making to have a successful build, I am also having to patch some ImGui-related variable names in `source/client/StarClientApplication.cpp` and `source/extern/imgui_iterator.inl`.

As I'm using ImGui 1.92.7 to build this I suspect some version mismatch so I'm holding off on including those changes.

I'm using GNU Make for compilation, after invoking CMake like this:
> cmake -Bbuild -Ssource/ -DSTAR_BUILD_QT_TOOLS=ON -DCMAKE_CXX_FLAGS="-D_DEFAULT_SOURCE -D_GNU_SOURCE" -DCMAKE_BUILD_TYPE=Release

### DIFF
--- a/source/CMakeLists.txt
+++ b/source/CMakeLists.txt
@@ -82,6 +82,8 @@ if(NOT DEFINED STAR_SYSTEM)
     set(STAR_SYSTEM "freebsd")
   elseif(${CMAKE_SYSTEM_NAME} STREQUAL "NetBSD")
     set(STAR_SYSTEM "netbsd")
+  elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Haiku")
+    set(STAR_SYSTEM "haiku")
   elseif(UNIX)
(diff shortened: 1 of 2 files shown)


Peppersawce:
Building with clang instead of gcc "fixes" the crashes I've been encountering on the OS.
I've created a new player and spawned inside the intro mission without any issues.

PR #506, opened 2026-04-20
Add initial Android support

lotigara:
Compiles & launches!
Crashes when connecting to a server.
Thanks to @RohanBhattacharyya for his [port](https://github.com/RohanBhattacharyya/oSBM)

TODO:
* [ ] Decide how to add touchcontrols
* [ ] Actually add it
* [ ] Build instructions
* [ ] Code quality
* [ ] etc.


J0w03L:
You may want to split this out a bit instead of doing everything as one commit. Right now, that single commit touches 325 files, adding nearly 164k lines. A lot of those look like they're external libs, but it is going to be very hard for anyone to review the work you've done here, and therefore it's going to be hard for anyone to trust that it is safe to merge.

lotigara:
Well, I'm closing this PR then. I'll split it later.

lotigara:
And a question: if I absolutely must use external libs, can they be included as Git submodules?

J0w03L:
External libs are fine to use, but a lib like that should be ported to vcpkg and included that way instead.

lotigara:
> External libs are fine to use, but a lib like that should be ported to vcpkg and included that way instead.

Should I also package libexecinfo? Or can I set minimal API to 34 (Android 14) where `execinfo.h` is already present?
**EDIT**: oSB with the `libexecinfo` crashes on every exception, even on non-critical ones (for example, empty server address). Android's execinfo is performing way better and in that case shows an error message.

lotigara:
~~NG-GL4ES is shim for OpenGL support on OpenGL ES-only systems, i.e. most Android devices. Though implementing OpenGL ES support is possible, but it's much easier to maintain code that uses a single API.~~

~~Unfortunately, that library is abandoned and therefore can't be included in vcpkg.~~

~~Is there another way to use that library, like using a submodule?~~
**EDIT**: the library seems to be active and had a release last month. I've opened an issue (https://github.com/BZLZHH/NG-GL4ES/issues/21) about vcpkg port

PR #507, opened 2026-04-20
Check for valid ContainerInteractor::openContainer results

J0w03L:
This PR removes the rather useless exception from `ContainerEntityPtr const& ContainerInteractor::openContainer()` in favor of simply checking if it is valid wherever it is used, which fixes a crash when interacting with a container at the same moment that the player walks out of its range.

### DIFF
--- a/source/frontend/StarContainerInteractor.cpp
+++ b/source/frontend/StarContainerInteractor.cpp
@@ -1,5 +1,4 @@
 #include "StarContainerInteractor.hpp"
-
 namespace Star {
 
 void ContainerInteractor::openContainer(ContainerEntityPtr containerEntity) {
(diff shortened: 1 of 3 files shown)


source/frontend/StarContainerInteractor.cpp
```diff
@@ -51,35 +48,67 @@ List<ContainerResult> ContainerInteractor::pullContainerResults() {
 }
 
 void ContainerInteractor::swapInContainer(size_t slot, ItemPtr const& items) {
-  m_pendingResults.append(openContainer()->swapItems(slot, items).wrap(resultFromItem));
+  auto container = openContainer();
```
Novaenia: this and the others below can be put under if (auto container = openContainer()) instead

Issue #204, opened 2025-03-01
Shaders jumpscare you with texture atlas when SSAA is enabled


Bottinator22:
SSAA spectacularly breaks shaders.

We/I need to either:
- Make SSAA and shaders compatible
- Disable SSAA when shaders are present and/or enabled

Bottinator22:
note: this is not a new issue, has been known for a while

PR #510, opened 2026-04-24
WIP fix shaders misbehaving with multisampling enabled

emmyposs:
largely complete, submitting as WIP to force myself to finish it (i have been sitting on this for 5 months Help)
if i dont update this in like 3 days yeel at me, throw bricks at me, etc

### DIFF
--- a/source/application/StarRenderer_opengl.cpp
+++ b/source/application/StarRenderer_opengl.cpp
@@ -235,7 +235,7 @@ void OpenGlRenderer::loadConfig(Json const& config) {
 
   for (auto& pair : config.getObject("frameBuffers", {})) {
     Json config = pair.second;
-    config = config.set("multisample", m_multiSampling);
+    config = config.set("multisample", pair.first == "main"? m_multiSampling : 0);
     Logger::info("Creating framebuffer {}", pair.first);
     m_frameBuffers[pair.first] = make_ref<GlFrameBuffer>(config);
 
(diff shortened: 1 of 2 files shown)


apnex:
Hi @emmyposs — I ran into this independently in a fork and ended up at the same diagnosis you did here, so I wanted to leave the confirmation on your PR rather than just open a competing one.

Reproduced it on hardware (Intel Arc / Mesa) by instrumenting the effect-texture bind site. `want` is the texture we asked to bind, `bound` is what the unit actually holds afterwards:

```
uniform=emission   unit=2  want=10  bound=10  err=0x0     <- a CPU-uploaded texture: binds fine
uniform=lightMap   unit=4  want=7   bound=0   err=0x502   <- a framebuffer texture: INVALID_OPERATION
uniform=lightMap   unit=4  want=7   bound=11  err=0x502   <- the bind is a no-op; the unit keeps its old binding
```

A CPU-uploaded sampler in the same frame binds cleanly — only framebuffer-sourced samplers fail. Over one session, with `antiAliasing` on in both legs:

| | before | after |
|---|---|---|
| `GL_INVALID_OPERATION` at the bind | 60,267 | 0 |
| mean world luminance | 0.016682 | 0.257642 |

So the failed bind is silent, the texture unit falls back to whatever was on it (an atlas page), and the shader samples that — which lines up exactly with #204's title.

I've opened #561 with the per-framebuffer opt-in, rebased on current `main` (this PR predates #542, which rewrote `switchEffectConfig`, so it no longer applies cleanly). It is substantially your change — you found this first and it should be credited to you. If a maintainer would rather bring this PR up to date and merge it instead, I'm happy to close mine; I mostly wanted the repro attached somewhere visible so it stops sitting unreviewed.

PR #516, opened 2026-05-03
post update function for player scripts, direct humanoid facing access & head rotation access

Bottinator22:
Useful for scripts that handle custom player animation.

This allows fixing the 1-frame delay between an item updating the player's facing and an update of `mcontroller.facingDirection` by using `player.facingDirection` instead and running the animation code in a generic player script's `postUpdate` function.

Direct humanoid facing access also allows for fixing issues relating to facing suppression when holding an item (it updates humanoid facing but mcontroller facing remains locked, resulting in `mcontroller.facingDirection` being incorrect)

I also added a callback for getting head rotation since that's probably going to be useful too.


Bottinator22:
I have no idea why the animated parallaxes commit is included considering that's... already merged.

PR #519, opened 2026-05-05
allow disabling tech overrides

Bottinator22:
allows disabling tech overrides with the `disallowTechOverrides` generic player property.

Useful for personal techs and for tech-reliant races like my own Drone and Worm races.


Novaenia:
this should probably be its own player.* function instead?

Bottinator22:
closing this, I'll probably totally rework it instead

PR #526, opened 2026-05-18
fix: improve biome foreground collision handling in WorldGenerator

Mofurka:
I love the collision modification for blocks, but there is little annoying thing with the planet terraforming. It replaces the collision of such blocks with a standard one

### DIFF
--- a/source/game/StarWorldGeneration.cpp
+++ b/source/game/StarWorldGeneration.cpp
@@ -30,6 +30,26 @@ namespace Star {
 
 static int const PlantAdjustmentLimit = 2;
 
+namespace {
+
+bool isPreservedTileCollisionOverride(CollisionKind collision) {
+  return collision == CollisionKind::None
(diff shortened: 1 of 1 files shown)


source/game/StarWorldGeneration.cpp
```diff
@@ -1029,7 +1054,7 @@ void WorldGenerator::prepareSector(WorldStorage* worldStorage, ServerTileSectorA
       replaceBiomeBlocks(tile);
       placeBiomeGrass(worldStorage, tile, position);
 
-      tile->collision = maxCollision(tile->collision, materialDatabase->materialCollisionKind(tile->foreground));
```
Novaenia, line 1032: maxCollision seems important?

Mofurka: Yep, maxCollision dont need here because it simply selects the stronger collision, which overrides manual overrides like platform or none with the new block material

Issue #242, opened 2025-05-16
Gamepad support?


DartPower:
Can you add gamepad support?

Shadowblitz16:
Is there a anyone working on this?

I would love to play this with my sister on batocera via controller and split screen like terraria on the switch.

Issue #298, opened 2025-07-20
[funtion request]controller support like xbox？


Sunbrella114514:

Shadowblitz16:
This game needs dedicated controller support. 
I want to use it on batocera linux but it's a controller focused system.

DarianLStephens:
I'd love to see it, primarily so I can play on the Steam Deck without resorting to the mouse all the time.

If it's possible to read controller inputs via LUA, then a controller-based interaction system could be scripted.

Sunbrella114514:
Since PR #530 has largely implemented the controller functionality, this issue can now be closed.

PR #530, opened 2026-05-25
Native Controller / Gamepad Support

neonspectra:
# Overview
This PR adds full gamepad support with analog stick input, multiple input modes, haptic feedback, and an in-game settings UI.

I've done my best to generally keep scope of changes limited to minimum necessary, but the unfortunate reality is that this is an incredibly low-level systems overhaul for a game that was never designed with controller gameplay in mind. This is a huge PR touches a lot. Various tweaks and plumbing refinements were implemented as required to make controller-based gameplay cohesive and paper over some nasty assumptions that were clearly made with a mouse and keyboard in mind.

It is worth noting that this PR does **not** implement a focus-based menu navigation system. The Starbound UI system is an absolute spaghetti mess and it would require a full overhaul to fix the menuslop in this game (much less do it in a way that accounts for FU and other mod-based UIs). Instead, this PR implements a virtual cursor (for trackpad-less gamepad users) and have a hybrid mouse mode for users whose gamepad has an integrated trackpad.

But, without further ado, here are the highlights:

# Highlights

#### Twin-Stick Aim!
*Not only is input fully omnidirectional 360° and magnitude sensitive, but with the hybrid input system you can even aim completely independently of the cursor, allowing you to use both analog sticks and a trackpad at the same time. Steam Deck users rejoice!*

demo2

#### Settings UI!
*Don't like the analog defaults? You can change them.*
controllersettings

#### Bindings menu!
*While this PR features a fully redesigned control scheme with context-sensitive bindings to maximise what you can do on your controller, most of the controls are rebindable. Unfortunately, context sensitivity is only accessible at the C++ layer, so there's no easy way to separate these outright at runtime without a full rebind menu overhaul. But you can always rebind away from them if you don't like them.*

modbindingscontroller

### Testing

Manually tested on Steam Deck (SteamOS 3.7, Desktop Mode) with Frackin Universe and 28 Workshop mods installed. Also tested with a DualSense (PS5) controller connected via Bluetooth. I don't have an Xbox controller (or any other controllers for that matter) so I didn't test that, but I have to hope that it works as well 😅 

Verified:
- General analog movement, aim, and basic gameplay with nothing obvious broken or unusable
- All four input modes including live switching via settings pane
- Virtual cursor & pane interaction (inventory, crafting, containers, starmap)
- Complex movement: platform drop-through, zero-g, swimming
- Mech deployment: aim origin, direction retention, ground impact rumble
- New context-sensitive overloaded actions: MM toggle/drop item, teleport combo, context-sensitive pane click overrides)
- Controller settings menu
- Frackin Universe compatibility (no crashes, no obvious broken UI)

---

*Technical notes from my lovely agent are below...*

---
---

# Technical Overview & Info for Agents

### Files Changed

**Core input** (`StarClientApplication.cpp/hpp`):
- Three active input modes (Auto/Gamepad/Hybrid) plus Off. Auto switches on last input device, Gamepad uses virtual cursor for menus, Hybrid decouples stick aim from mouse cursor for trackpad devices.
- Left stick: analog movement via `setMoveVector()` with deadzone normalization. Vertical axis triggers `moveUp()`/`moveDown()` past configurable threshold for zero-g, swimming, and platform drop-through.
- Right stick: world-space aim offset relative to player/vehicle position. Direction retained on release. Vehicle aim origin uses the vehicle entity's position, not the player anchor point.
- Trigger axis→button conversion (threshold 0.5) for LT/RT since the bind system only handles `ControllerButton`.
- Active controller filtering — locks to first device with significant input, ignores others. Handles Steam Deck's dual gamepad registration (physical Neptune + Steam Virtual Gamepad). 120-frame idle timeout for controller handoff.
- Virtual cursor: right stick drives screen cursor with synthetic mouse events through `MainInterface::handleInputEvent`. All existing panes/widgets work unmodified.
- R3 right-click when panes are open (stack splitting, star map travel). Camera pan suppressed during.
- Rumble on damage (intensity scales with fraction), beam teleport, and vehicle ground impact (uses `Vehicle::onGround()` transition + fall velocity).
- Weapon crosshair cursors suppressed when controller aims. Interaction/inspect cursors preserved.

**Bindings** (`controller.binds`):
- New `.binds` file with `"controller"` category. Entries match `InterfaceAction` names so `isActionTaken()`/`isActionTakenEdge()` bridge both legacy keyboard binds and controller binds. Existing `bindings` config and mod compatibility preserved.
- Default layout: RT/LT fire, shoulders cycle hotbar, face buttons for jump/interact/inventory/close, d-pad for hotbar group/MM toggle/quest/tech, LB+RB beam combo, context-sensitive L3/R3.
- Standalone `ShiftDropAction` bind (unbound by default) for users who want MM single-block mode / drop on a dedicated button regardless of input mode.
- All binds appear in existing Mod Binds menu and are rebindable to any controller button.

**UI** (`StarControllerSettingsPane`, `StarOptionsMenu`):
- Controller button in Options menu opens a `BaseScriptPane` subclass with `LuaRoot` for `root.getConfiguration`/`root.setConfiguration` access.
- Mode selection (Off/Auto/Gamepad/Hybrid), sliders for deadzone, aim range, cursor speed, drop threshold. Settings apply live via per-frame config re-read.

**Other touched files**: `StarInputEvent` (TriggerLeft/Right enum), `StarMainInterface` (setOverrideAim, cursor suppression), `StarPlayer` (isFiring), `StarVehicle` (onGround), `StarApplicationController` (rumble/rumbleTriggers/activeControllerName virtuals), `StarMainApplication_sdl` (SDL3 rumble impl).

### What it doesn't change

- Legacy `KeyBindings`/`InterfaceAction` system untouched. Keyboard binds in `starbound.config:bindings` are not affected.
- No changes to the `opensb.binds` file (controller binds moved to their own separate `controller.binds`).
- No Lua API changes that would affect mods.

See newly created documentation file `doc/CONTROLLER_SUPPORT.md` for architecture details, rationale behind design decisions, and implementation footguns.


neonspectra:
This PR fixes #242 and #298. Fixes the controller part of #3 (but sorry, no splitscreen 😅)

emmyposs:
blatantly ai generated

Bottinator22:
Some issues I spotted:
- the `dirOffset` value mentioned in the readme shouldn't be hardcoded for vehicles. Vehicle size can vary a lot. If vehicle size is being problematic, you should probably determine the value dynamically by the vehicle's collision poly. The current approach goes back to failing with anything bigger than a modular mech. There are a lot more vehicles than just mechs.
- Some of these numbers should be read from a config instead of hardcoded. Then if they aren't good enough, they can be changed with mods or settings.

- Did you test how this acts with a non-centered camera?
Here's a command that can be used to test this (hover over any entity to focus)
`/run player.setCameraFocusEntity(world.entityQuery(player.aimPosition(),1)[1])`

If you use `Player::cameraPosition`, that would likely give you a much more accurate aim reference than the current approach with no need for the special cases.

- Hardcoding rumble behaviour is... bad. The health-based rumbling and teleport rumbling should be handled by a player script of some kind. The vehicle landing rumble should be handled either by the vehicle script or at least the aforementioned player script. Not every vehicle is a mech, and not all vehicles land as hard as the Frackin Universe mechs.

...though considering this is AI generated, there are probably a lot more issues than what I listed here.
Do not AI generate Starbound code. You're going to spend more time and effort fixing the AI's mistakes just to get something you don't even understand.

grbr404:
So I'm going to be honest, I don't like how this PR and the changes are almost fully written by an agent (Ai).

Also, try to make the UI at least match OpenSB's UI.

Obviously, this is a decision to be made by the maintainers, but I don't think this should be merged regardless of the controller support.

neonspectra:
@Bottinator22 
> the dirOffset value mentioned in the readme shouldn't be hardcoded for vehicles. Vehicle size can vary a lot. If vehicle size is being problematic, you should probably determine the value dynamically by the vehicle's collision poly. The current approach goes back to failing with anything bigger than a modular mech. There are a lot more vehicles than just mechs.

I actually did an analysis of all the different vehicles in the game to come to this value. Even on the widest vehicles (the shuttles), the centrepoint is calculated based on the absolute vehicle centre. Using the value I chose here was an intentional choice because it didn't seem appropriate to add complexity as well as change the centring point for the return from something less consistent.

Is there any reason to change it? Who is going and making xboxhuge vehicles? Is the added complexity of adding an extra lookup here worth it? Like, you're not wrong that this isn't scaling but the big question is whether it makes the codebase better if we do so. WDYT? I can take another look if you really think so.

> Some of these numbers should be read from a config instead of hardcoded. Then if they aren't good enough, they can be changed with mods or settings.

Please advise on specific numbers or values you would like to see changed and the reasoning behind why they should be exposed as public APIs. My thoughts on this is that not everything needs a script hook IMO; this PR builds low-level input translation and largely just extends the C++ APIs that already exist. This is almost all code that needs to be fast, that needs access to low-level engine APIs, and just generally doesn't offer much value to expose more than what's already being exposed via the bindings menu and controller settings menu.

The biggest names here are going to be the bindings, and the controller sensitivity/drop threshhold stuff, which is in fact user configurable precisely because users may want to configure them.

> Did you test how this acts with a non-centered camera?
Here's a command that can be used to test this (hover over any entity to focus)
/run player.setCameraFocusEntity(world.entityQuery(player.aimPosition(),1)[1])

Had no idea you could do this. But I just tried it and I confirm that it works because we are aiming relative to the player/lounging entity. No issue, aiming works as expected in this situation.

> If you use `Player::cameraPosition`, that would likely give you a much more accurate aim reference than the current approach with no need for the special cases.

No, this isn't correct (and for the same reason that your above suggested test actually works correctly in the current implementation). We explicitly do not want to use `Player::cameraPosition` because that would break aiming. 

When you have twin-stick controls, everything is a circle around the player. If we have aim not be attached to the player, then you're always going to be aiming toward the camera centrepoint rather than aiming relative to your character (as you'd expect from any other twin-stick sidescroller).

> Hardcoding rumble behaviour is... bad. The health-based rumbling and teleport rumbling should be handled by a player script of some kind. The vehicle landing rumble should be handled either by the vehicle script or at least the aforementioned player script. Not every vehicle is a mech, and not all vehicles land as hard as the Frackin Universe mechs.

So hear me out please. This is tricky because the rumble API in SDL3 isn't natively exposed to scripts or even used at all in the engine as-is. My thought on this is that if we want to design a proper API for rumble modding, it's outside of the scope of this PR. We can bikeshed about rumble or we can implement the basic obvious QoL cases for it; I chose to go the direction of the former simply bc I think it's fun and I want it. 

If you want me to rip out those specific commits and push rumble off entirely to its own thing then sure, fair enough. Let me know if you think that's preferable.

@grbr404 
> Also, try to make the UI at least match OpenSB's UI.

So, the thing about this is that Starbound uses completely baked-in raster backgrounds and baked in textfor all of its UIs. I literally would have to open up GIMP and start fiddling with pixels to clean this up.

If this is the make or break that's the absolute showstopper that stops this PR from being merged then ok I'll do it but honestly I just want to play this game with a controller. i dont think that having yet one more slightly different UI on top of all the others we already have for this one niche thing is a real problem for basically anyone. 

Alternatively, if you wanted to contribute a new asset for this I'd be super appreciative 😊 

---
re: the anti-AI sentiment from several of you

I am not here to get into an ideological argument. I'm not ashamed to admit that I could not have created this PR without AI. I love my agent dearly and I don't think it adds much to the conversation the mention this when it's the most obvious thing in the world from the commit history that all this is agent-written. 

From where I am standing, the simple fact is that OSB has been around since, what, 2023(?) and in that time none of you have implemented this much-requested feature yourselves. I want to play Starbound on my couch and ghost damnit I'm going to play Starbound on my couch! I'd like to share the ability to do so with anyone who wants it! 🙂 

If you would like to implement controller support yourselves, you are very welcome. I am here to get shit done and that's as far as I'm going to say about that topic. I'm happy to continue working with you to get this PR to the finish line so please let's stay on topic and let's stay civil.

Bottinator22:
> Who is going and making xboxhuge vehicles?

ME. Also, I'm pretty sure XS Mechs has some pretty big mechs that are bigger than shuttlecrafts. Someone's going to make a bigger mech. Anything you add to oSB should be able to account for _everything._

> This is almost all code that needs to be fast

This is code that runs on update in StarClientApplication.cpp. It runs once per frame. It doesn't need to be hyper-optimized. We do need more script hooks for things, because script hooks allow mods to make changes. If someone doesn't like how something works in oSB, a script hook or config allows people to change those values with a mod.

> But I just tried it and I confirm that it works

What if you're supposed to aim relative to that entity? If the camera focus entity is far enough from the player to have the player offscreen, if you're using mouse aiming, the aim position will always be on-screen. But with your method, it looks like you can end up with an offscreen aim position in that case. This is problematic if a mod adds something like a drone item, something like the racer drone from Terraria.

> This is tricky because the rumble API in SDL3 isn't natively exposed to scripts or even used at all in the engine as-is.

So expose it to scripts. Add a Lua API for rumble support. Maybe add it to the input bindings. Then mods can decide when the rumble should happen.

> I literally would have to open up GIMP and start fiddling with pixels to clean this up.

I've done that for my options menu update for HDR. If you don't do it, you're making someone else do it later. I might be partially guilty of this too.

neonspectra:
Ok, nailed the vehicle aim offset scaling as a quick fix for now. That's resolved. Tested and verified it works.

---

For `Player::cameraPosition` - I did the fix you recommended and the issue I brought up on Discord is now happening. See below:

https://github.com/user-attachments/assets/34e4304f-c287-4f0e-93e3-b860988232ed

I'm reverting commit `15f14205df750132612decc9980e8e14323c0c3f` so that we're back to a working state. We NEED to track player for this. 

My original implementation works when you test against it with `/run player.setCameraFocusEntity(world.entityQuery(player.aimPosition(),1)[1])`, so it's the most robust version of the implementation for now.

> This is problematic if a mod adds something like a drone item, something like the racer drone from Terraria.

Is there a specific mod that exists that I can test against beyond doing the above command to 100% confirm the above?

---

For remaining two items, I have a to-do list. Not gonna get these tonight but I will investigate more thoroughly and get back to you when I have more on these:
- Rumble: Will investigate exposing an API for this and changing these bindings to be mediated by Lua. I want to have some time to think about this a bit before making the change, but I really do want to have rumble so I'll see how I can clean this up.
- Menu screen: Adding to my to-do list.

If anything else comes up in meantime, feel free to add to the list.

neonspectra:
Reverted version works: 

https://github.com/user-attachments/assets/cef0eb69-4b07-48ee-8302-6189ec6c0074

So remaining to-do list:
- Rumble fixes / exposing public API and moving this stuff to be configured by Lua.
- Menu screen asset

grbr404:
@neonspectra I can look into making the UI look correct for you. Just to say, I was mostly opposed to this PR not because it's written by AI, but rather because it's hard to know if the person using the AI actually understands the code. Based on your previous comments, I'm giving you the benefit of the doubt and assuming you know what you're doing.

neonspectra:
@grbr404 Thanks. You don't have to; I can get it done if you do not get to it. It may be easier for me to do it since I can move around the menu UI and whatnot when it comes to that fiddly pixel alignment stuff.

FWIW, I'm the one in the wrong on this ultimately and I should know better when it comes to this menu asset polish work. I think I was being too hasty with my initial PR and letting myself getting a little sloppy on the corners after spending the better part of the last two days playing whack-a-mole with edge cases on the actual functional parts of the code and letting that get to me.

And tbh Bottinator is right that OSB has to account for anything and then remain stable no matter what crazy stuff modders do, so we might as well take the time to make this PR a good experience since as soon as we merge modders will do what they will and we risk breaking mods if we have to revert or change something that was put in haphazardly.

I think that, again, fair enough not wanting slop in the project. It makes me happy that even after over a decade now people still give a shit about this game and feel strongly about the standards for contribution. Literally every open source project ever (including the ones where people actually get paid) is trying to fight slop PRs right now as a massive waste of maintainer time, so I get being wary and dismissive gauging whether someone's just trying to chuck garbage over the proverbial fence.

grbr404:
@neonspectra Don't worry, making UI is quite easy. It's still a work in progress and looks a bit rough right now because I'm actively editing it.
wip

neonspectra:
~~Another note to self: it seems that keyboard bindings (IE, mouse and keyboard, not controller bindings) are triggering their bound actions while I type in the game chat.~~

~~I don't normally use the chat, and my `controller.binds` adds exclusively gamepad bindings, so I have no idea if this is my bug or if it's a totally unrelated bug in upstream (or if it's something a totally unrelated mod is doing).~~

~~I need to investigate this and at the very least know + be able to explain what's going on here (and, ideally, throw in an unrelated fix if it's something pre-existing in upstream).~~

**EDIT:** This was my bug. When I extended `MainInterface::handleInputEvent` to let controller inputs interact with panes, it created a pathway for unrelated keyboard inputs from `isActionTakenEdge()` to be read unsuppressed. Added a guard to prevent this. Tested and verified the fix works. Resolved in commit directly below. Now onto other things.

neonspectra:
While we are talking about UI, I also made another UI fix for the **Controller** button itself within the options menu-- it now uses the correct sized button and is properly aligned in the menu:
optionsmenu

neonspectra:
That said, followup to the above: I was kinda rolling around in my head where the best spot to put this is. There is actually some space it could go above "Mod Binds", but then it's cramping the space below volume sliders. What do you guys think? Here are two alternate version mockups:

#### Option 2 - Mid-buttons left aligned
optionsmenu2

#### Option 3 - Mid-buttons centre aligned
optionsmenu3

grbr404:
This is how the controller UI now looks on my end. I have added `Rumble Intensity` as another setting that should be implemented as a multiplier that basically goes from 0 to 2, default 1, to allow users to tweak rumble to their preference.wip5

 And I think the buttons for mode select should work like the select buttons in the shader menu:s
Currently, they are just normal buttons, and I would appreciate it if you could implement that, as I am not that good at Lua.

neonspectra:
@grbr404 I really like this. As for buttons change, yes I can do that. Should not be an issue.

Did you implement rumble intensity or is it a mockup slider on your end? Either way is fine since implementation of rumble needs a cleanup overhaul to expose it to Lua for letting modders attach it to stuff. Anything that is currently mockup can be fixed on my end to match the above.

Can you PR my branch ([`neonspectra:feature/controller-support`](https://github.com/neonspectra/OpenStarbound/tree/feature/controller-support)) with the asset and/or any changes you made to `controller.config` or the game code so it can get pulled into this PR? Alternatively, I sent you collaborator on my fork if you find it easier just drop it right into the PR that way; either way works for me.

grbr404:
@neonspectra I will make a PR.

neonspectra:
@grbr404 Oh whoops, I see we both made the changes at the same time. I saw your comment with the zip file assets before you deleted it lol, but all good. Also TIL for the being unable to fork a fork thing; I also had no idea that was a limitation.

My change also wired the thing up to Lua, but it's still in mockup state as the C++ is hardcoded. I will go ahead and rebuild on my end to see what we're working with and make the UI changes to properly work like the mockup and factor the config option into the rumble rework. Thank you so much for making this UI for this PR 😁

grbr404:
@neonspectra Sorry for the mess, by the way. Just one thing I noticed: the active mode should probably be lit up green, like I showed with the shader menu example.

If any other UI stuff comes up, I will just attach them to a comment instead.

grbr404:
@neonspectra This is how the Select button is implemented in the shader menu. I think that the controller menu should match that.
https://github.com/OpenStarbound/OpenStarbound/blob/fc3d55ebbbecc5600a21d3e9e95ae06cf5081f9e/assets/opensb/interface/opensb/shaders/shaders.lua#L154-L178

neonspectra:
@grbr404 So before even getting into the button selection style change, I rebuilt my client with your assets and it looks like I'm running into a scaling issue here.

Below is what I get running this using the png assets and `controller.config` you added in [`fcab79587264a4410edf56e45400dab5379b77c4`](https://github.com/OpenStarbound/OpenStarbound/pull/530/changes/fcab79587264a4410edf56e45400dab5379b77c4):

Screenshot_20260526_210300

My commit after doesn't touch anything that should relate to scaling these assets, but just to be absolutely sure I rebuilt my assets with the git tree set to that above named commit.

What's interesting here is that I looked through the `interfaceScale()` function and found that background images and widget buttons are multiplied by the same factor for interface scaling, so in theory there shouldn't really be any scaling issues from the client itself. 

The only thing I can think of is do you have any client mods scaling your UIs differently? The reason I ask is while I could just scale the assets up and call it a day, the fact that there's a discrepancy between two different people is concerning.

grbr404:
@neonspectra It seems like I accidentally overwrote my updated version with the one prior to the Rumbel intensity slider before uploading. Just replace it with the one attached.
body

neonspectra:
@grbr404 Perfect, thank you!! Replaced the file, added in the button code from the shader menu as a reference, and verified it works. This looks fantastic! Amazing work! 💖

https://github.com/user-attachments/assets/bb6cc574-c342-4d7a-80b8-eb5d45e9eed2

worth noting: rumble slider still not in use at current commit state. i will clean this up with the rumble rework. but for now it is eppy eepy bed by time.

neonspectra:
Fixed support of the title screen, so that works now. Tested and verified. 

Next up will look into rumble and will do another pass looking over what could/should be exposed to Lua.

Sunbrella114514:
I’m positive about this PR.

You and I think alike. Ever since I switched from the Steam Deck to the ROG Ally, I’ve been wishing Starbound had proper controller support like the Xbox version, so I could enjoy my little universe more comfortably on a handheld.

As for AI, my view has always been the same as the developer above: if you know what you’re doing, then go for it.

I’m looking forward to your work, and I also respect other developers’ opinions on AI.

Sunbrella114514:
消息
After playing through it for a while, I wanted to share some issues and suggestions I encountered:
Gun aiming feels quite awkward without a visible cursor.
Maybe you could try adding a laser sight or aiming beam.
For the Matter Manipulator, you could take inspiration from the Xbox version by adding a highlighted mining outline based on the mining area size.
The current highlight feels less noticeable because there is no visible cursor.
I would recommend changing the Shift key function to L3, and moving cursor toggle to R3, similar to how most shooter games handle sprinting.
(I also could not seem to find a binding for the Shift key.)
Please allow key bindings for joystick directional inputs as well.
For example, I would like to bind pushing the left stick downward to crouching.
It would also be nice if inventory-related interfaces had item slot highlighting, or quick item transfer through combination key shortcuts.
Overall though, despite these issues, the experience is genuinely very enjoyable.
By the way, if you are okay with it, could I ask the OSBM developer to reference your implementation or code approach? I think your controller support design fits mobile platforms extremely well.

neonspectra:
@Sunbrella114514 TY for the feedback! Need to look into some of these more, but here's my off the cuff responses to the items you brought up so let me know your thoughts.

- `Gun aiming feels quite awkward without a visible cursor.` Very valid opinion, but this is tricky because of a couple things:
    - How should we handle different weapons? 
        - Should there be one graphic for all of them? If we want per-weapon differences, then this opens a really big can of worms because it requires heavily touching all the base game files to assign a graphic and then also exposing something to Lua to let modders change this for custom items, and also account for mods that don't set this. I am hesitant and leaning toward saying that this would be only viable whatsoever if there is one graphic across the board for everything no matter what
    - How do we make it feel non-immersion breaking? If we have a laser sight that gets stopped by walls, this again turns into a really big overhaul because then we have to touch collision code and add new elements there. Not necessarily the worst idea, but also yeah, idk. Also this kinda feels out of place for things like bows, anyway, so i sorta lean against this
    - but also, having just a boring straight line on the screen seems like it would look bad / feel hacky or debugy and immersion breaking. What do you think?
    - my personal gut feeling on this is that both the simplest route and the old school game development approach is to just leave it cursorless and consider it just a skill for the player to practice as part of the experience of playing the game.
- `For the Matter Manipulator, you could take inspiration from the Xbox version by adding a highlighted mining outline based on the mining area size.`
    -  This seems reasonable. I will look into it. That said, this is one of those awkward pre-baked mousebrained assumption sins the original game makes since there is no highlight in the base game so this is something I'll need to come up with from scratch (I don't have a console so I don't know what the xbox version looks like). I'll see what I can do and see if there's a way to make it look not terrible.
- `I would recommend changing the Shift key function to L3, and moving cursor toggle to R3, similar to how most shooter games handle sprinting.` 
    - Breaking down the three uses of shift in the game (that I know of):
        - Matter manipulator mining size: shift is actually set up this way contextually in hybrid mode. i think i've accidentally without thinking made a self report that i never actually use controllers without a trackpad so the virtual cursor thing was kind of an afterthought 😅 
        - Walking: Controller movement is magnitude based (so you can walk by moving the stick less), so this functionality isn't needed here and should not take up a precious control slot IMO
        - Moving stack between inventories quickly: Genuinely overlooked this. I will try to think about what can be done here. Maybe this needs a contextual binding, idk. This is hard.
    - so, the biggest problem with coming up with a default layout right now is that there's no concept of "per-mode" defaults (and it would be super complex and not really benefit anyone to add them since you only have defaults once), so we need one overarching set of default controls.
    - The way I've handled it as-is to have a unified set of defaults is that the binding for virtual cursor simply does something different in hybrid mode (since the whole point of hybrid mode is that there is no virtual cursor) and included that same control as a separate bindable option for those who use normal gamepad mode to rebind where they see fit.
    - however, i still think it would be good before merging for there to be a final pass on what the defaults should be to get more opinions on this. that said, i literally have no idea what the controls for this look like on the official port since i do not have a console and even googling this does not get me results as to what they are. this information seems to be so basic for most people that it doesn't actually exist on the internet for those who do not have this version of the game.
        - that said, if the official port has a focus system then we inherently have to be different anyway because we have a virtual cursor system here, which leads me to...
- `It would also be nice if inventory-related interfaces had item slot highlighting, or quick item transfer through combination key shortcuts.`
    - So I thought about this and the biggest problem here is that it requires an entire UI system overhaul to do this, and kind of goes against the OSB mentality of trying to keep changes both minimal and backward compatible for modding purposes.
    - The reality is made even more complicated by the face that many mods add their own custom UIs, so making focus-based controls work in such an environment and predict anything anyone is going to do is pretty much impossible. there's no world in which we can avoid having a virtual cursor for gamepad users, so it seems to me like if that's something that someone wants to build down the line, it should be scoped as a separate additive PR after controller support is already here.
- `Please allow key bindings for joystick directional inputs as well.`
    - This is tricky because analog sticks are by defninition not buttons. They give you vector and magnitude, so there is no concept of a button press and we intentionally make use of them exclusively for twin-stick movement because that's the entire point of controller support.
    - `For example, I would like to bind pushing the left stick downward to crouching.` -- My implementation, I just treated crouching as a control that simply didn't make the cut for being important enough to be bound to one of the limited buttons that we have available. But there's a couple of potential options for this. What do you think of any of the below:
        - There could be a binding for crouch if a player, say, wants to bind a back button to this or something?
        - We could look for a percentage of magnitude in a specific vector range and crouch if we reach this and character is standing? Unsure if it makes sense to make this a toggleable option with a crouch button alternative?
 - `By the way, if you are okay with it, could I ask the OSBM developer to reference your implementation or code approach? I think your controller support design fits mobile platforms extremely well.`
    - oh yeah i dont believe in copyright so do whatever with my code. information wants to be free. :)
    - given the nature of the OSB project copyright... situation... consider this comment to be my formal signoff that my contribution is public domain and can be used or referenced by anyone for any purpose

astyrah:
@neonspectra 

for the controller/weapon aiming, maybe you can do it like terraria does aiming for controllers? where an aim cursor (circle) shows up temporarily where the joystick region is. no lasers or tracers just a crosshair/indicator where the current aim is pointing. you can try playing terraria to get an idea of what i mean

i do agree however with keeping it simple (best aesthetically) and no lines/lasers as it would be awkward for many weapon types (plus having to align it to every gun's barrel for aesthetics). leave it to player's skill for aiming. --or-- at the very least make it an option to not have tracers/lasers for people who dont want it. having options is the best for people

Sunbrella114514:
> @Sunbrella114514 TY for the feedback! Need to look into some of these more, but here's my off the cuff responses to the items you brought up so let me know your thoughts.
> 
> * `Gun aiming feels quite awkward without a visible cursor.` Very valid opinion, but this is tricky because of a couple things:
>   
>   * How should we handle different weapons?
>     
>     * Should there be one graphic for all of them? If we want per-weapon differences, then this opens a really big can of worms because it requires heavily touching all the base game files to assign a graphic and then also exposing something to Lua to let modders change this for custom items, and also account for mods that don't set this. I am hesitant and leaning toward saying that this would be only viable whatsoever if there is one graphic across the board for everything no matter what
>   * How do we make it feel non-immersion breaking? If we have a laser sight that gets stopped by walls, this again turns into a really big overhaul because then we have to touch collision code and add new elements there. Not necessarily the worst idea, but also yeah, idk. Also this kinda feels out of place for things like bows, anyway, so i sorta lean against this
>   * but also, having just a boring straight line on the screen seems like it would look bad / feel hacky or debugy and immersion breaking. What do you think?
>   * my personal gut feeling on this is that both the simplest route and the old school game development approach is to just leave it cursorless and consider it just a skill for the player to practice as part of the experience of playing the game.
> * `For the Matter Manipulator, you could take inspiration from the Xbox version by adding a highlighted mining outline based on the mining area size.`
>   
>   * This seems reasonable. I will look into it. That said, this is one of those awkward pre-baked mousebrained assumption sins the original game makes since there is no highlight in the base game so this is something I'll need to come up with from scratch (I don't have a console so I don't know what the xbox version looks like). I'll see what I can do and see if there's a way to make it look not terrible.
> * `I would recommend changing the Shift key function to L3, and moving cursor toggle to R3, similar to how most shooter games handle sprinting.`
>   
>   * Breaking down the three uses of shift in the game (that I know of):
>     
>     * Matter manipulator mining size: shift is actually set up this way contextually in hybrid mode. i think i've accidentally without thinking made a self report that i never actually use controllers without a trackpad so the virtual cursor thing was kind of an afterthought 😅
>     * Walking: Controller movement is magnitude based (so you can walk by moving the stick less), so this functionality isn't needed here and should not take up a precious control slot IMO
>     * Moving stack between inventories quickly: Genuinely overlooked this. I will try to think about what can be done here. Maybe this needs a contextual binding, idk. This is hard.
>   * so, the biggest problem with coming up with a default layout right now is that there's no concept of "per-mode" defaults (and it would be super complex and not really benefit anyone to add them since you only have defaults once), so we need one overarching set of default controls.
>   * The way I've handled it as-is to have a unified set of defaults is that the binding for virtual cursor simply does something different in hybrid mode (since the whole point of hybrid mode is that there is no virtual cursor) and included that same control as a separate bindable option for those who use normal gamepad mode to rebind where they see fit.
>   * however, i still think it would be good before merging for there to be a final pass on what the defaults should be to get more opinions on this. that said, i literally have no idea what the controls for this look like on the official port since i do not have a console and even googling this does not get me results as to what they are. this information seems to be so basic for most people that it doesn't actually exist on the internet for those who do not have this version of the game.
>     
>     * that said, if the official port has a focus system then we inherently have to be different anyway because we have a virtual cursor system here, which leads me to...
> * `It would also be nice if inventory-related interfaces had item slot highlighting, or quick item transfer through combination key shortcuts.`
>   
>   * So I thought about this and the biggest problem here is that it requires an entire UI system overhaul to do this, and kind of goes against the OSB mentality of trying to keep changes both minimal and backward compatible for modding purposes.
>   * The reality is made even more complicated by the face that many mods add their own custom UIs, so making focus-based controls work in such an environment and predict anything anyone is going to do is pretty much impossible. there's no world in which we can avoid having a virtual cursor for gamepad users, so it seems to me like if that's something that someone wants to build down the line, it should be scoped as a separate additive PR after controller support is already here.
> * `Please allow key bindings for joystick directional inputs as well.`
>   
>   * This is tricky because analog sticks are by defninition not buttons. They give you vector and magnitude, so there is no concept of a button press and we intentionally make use of them exclusively for twin-stick movement because that's the entire point of controller support.
>   * `For example, I would like to bind pushing the left stick downward to crouching.` -- My implementation, I just treated crouching as a control that simply didn't make the cut for being important enough to be bound to one of the limited buttons that we have available. But there's a couple of potential options for this. What do you think of any of the below:
>     
>     * There could be a binding for crouch if a player, say, wants to bind a back button to this or something?
>     * We could look for a percentage of magnitude in a specific vector range and crouch if we reach this and character is standing? Unsure if it makes sense to make this a toggleable option with a crouch button alternative?
> * `By the way, if you are okay with it, could I ask the OSBM developer to reference your implementation or code approach? I think your controller support design fits mobile platforms extremely well.`
>   
>   * oh yeah i dont believe in copyright so do whatever with my code. information wants to be free. :)
>   * given the nature of the OSB project copyright... situation... consider this comment to be my formal signoff that my contribution is public domain and can be used or referenced by anyone for any purpose

Thank you very much for your reply!

First of all, regarding the Xbox version: although neither of us has actually played it, there are fairly complete gameplay videos on https://m.youtube.com/watch?v=QE4pCS3fR_k&t=721s&pp=ygUOc3RhcmJvdW5kIHhib3g%3D that you can use as reference. It is honestly impressive that you managed to build such a complete controller system even without direct reference material.

About the cursor issue:

The Xbox version seems to rely on some kind of auto-aim / target assistance system, which avoids the problem of invisible aiming. However, proper aim assist is probably quite complicated to implement, and I am not even sure it would necessarily feel better than Terraria-style controller assistance.

You are also right that a visible straight aiming line could hurt immersion. A “real” laser sight would also be difficult to implement properly. Maybe manually adjusting the aiming angle could become an interesting design choice, although personally I think it might become tiring over long play sessions. Controller gameplay usually gives people the expectation of additional aiming assistance.

About the Matter Manipulator highlight:

You could probably reference the Xbox version for this as well (the gameplay video linked above). I am not sure whether mods expose enough APIs for implementing such an outline, or whether rendering it client-side would become too bloated.

About Shift:

I actually forgot that Starbound does not really rely on Shift for sprinting, so that key is probably not very important after all.

About the default layout:

I think many buttons could support multiple contextual functions. For example, the Alt key (free look camera) could also double as a quick-loot button for containers, since players usually do not need free camera movement while interacting with storage containers anyway.

About crouching:

I do not think toggle crouching would fit Starbound very well. Unlike many other games, crouching is not really a persistent character state, because movement immediately cancels it. Mapping crouch to pushing the stick downward is mainly intended to replicate how the S key works under WASD controls, while also freeing up R3 for something more useful.

I also noticed something quite interesting: although the analog stick properly supports character movement, Starbound’s walking animations clearly were never designed for extremely small movement amplitudes. I wonder how the Xbox version handled this. Maybe they simply mapped stick directions directly into WASD-style 8-direction movement.

About code licensing:

If you are okay with it, I would like to open an issue on the OSBM side and ask Rohan whether he would be interested in referencing your controller implementation.

I am currently using Claude to help integrate and adapt code, and I have already tested your controller implementation on mobile devices. I have been considering whether I should send over some of the modified code I worked on. If you are comfortable with that, I would be extremely grateful, although I understand it may be a somewhat intrusive request.

Finally, thank you again for your work. You accomplished one of my three long-time wishes since I first started playing Starbound.

astyrah:
@Sunbrella114514 
> About Shift:
> 
> I actually forgot that Starbound does not really rely on Shift for sprinting, so that key is probably not very important after all.

iirc shift on mouse/keyboard is used for precision building (1x1 block placement/removal) and walking. so it has it's uses as a keybind (more on the precision and less on the walking unless RPing)

neonspectra:
@Bottinator22 Ok, I'm going through the rumble improvements and exposing the damage and beam up/down to Lua is pretty simple, but vehicle landing is a whole different can of worms. I want to have this because the mech landing feels powerful and it's quite simply (at least in my opinion) the most obvious candidate for where a rumble feedback feels great.

TL;DR is that the code that does the landing detection lives serverside, but rumble for obvious reasons needs to be a clientside function and the way the script callbacks are currently written don't have a clean way of bridging the gap. So we have to make sort of unrelated helper code that would otherwise be outside the scope of this PR to make that functionality happen outside of C++ if we want it.

---

To go into detail about the problem:

The landing detection requires checking the vehicle entity's `onGround()` state (from its `MovementController`). In C++ we do this by looking up the vehicle via `player->loungingIn()` → `worldClient->entity(entityId)` → `as<Vehicle>()` → `vehicle->onGround()`. But:
1. Vehicle scripts (`StarVehicle.cpp` line 192-197) get `vehicle`, `config`, `entity`, `mcontroller`, and `animator` callbacks, but no input callbacks. They can detect their own landing (the mech Lua script already does at `mech.lua:271` with `mcontroller.onGround()`), but they have no way to call `input.rumble()` since that API lives client-side and vehicle scripts run in a shared client/server context without access to the input system.
2. Player scripts have input callbacks (via `UniverseClient::setLuaCallbacks("input", ...)` at `StarClientApplication.cpp:777`) and can detect they're lounging via `player.loungingIn()`. But there's no Lua-exposed way for the player to query the vehicle's `MovementController::onGround()` state. The player's own `mcontroller` is inactive while lounging; the vehicle's movement controller is driving physics, not the player's.

---

The way I see it, there's two primary viable options here to add a Lua callback and I'd like your input before I go and throw this at my agent to implement:                      

- Option 1: **`world.entityOnGround(entityId)`** — Add this to `StarWorldLuaBindings.cpp` alongside the existing `world.entityPosition()`, `world.entityVelocity()` pattern. Implementation looks up the entity by ID, then         
 dispatches by type: `Vehicle` → `vehicle->onGround()` (the public accessor this PR already added to `StarVehicle.hpp`), `Player`/`Npc`/`Monster` → their `ActorMovementController::onGround()`. Returns `nil` for entity types without a movement controller or invalid IDs. This is the most general-purpose; modders can query ground state for any entity, not just the player's vehicle.                                                             

- Option 2: **`player.vehicleOnGround()`** — Add this to `StarPlayerLuaBindings.cpp`. Internally does `player->loungingIn()` → looks up the vehicle entity → calls `vehicle->onGround()`. Returns `nil` if not lounging. Narrower scope, fewer lines, less unrelated code touched, but only useful from the player script context and only for the player's current vehicle.                                                                                                      

In both cases the new callback is one function registration + ~10 lines of implementation. Neither implements changes to existing APIs. The player rumble script would then do something like:                                               

 ```lua
   -- in update()
   local grounded = world.entityOnGround(loungeId) -- or player.vehicleOnGround()
   if grounded and not self.lastGrounded and self.lastYVel < -10 then
     input.rumble(intensity, intensity * 0.7, duration)
   end                                               
 ```

Doing option 1 follows existing `world.entity*` conventions better, but option 2 is more minimal. 

I personally sort of lean toward option 1 because it would give modders more ability to hook into the `onGround()` check if they implemented their own custom stuff that they wanted to add rumble to (or use this check for other reasons). WDYT?

Bottinator22:
See LuaEntityUserdata.cpp, there are existing mcontroller bindings there
that currently don't support non-actor mcontrollers

Sunbrella114514:
［Issue］When using a pickaxe in controller mode, there seems to be neither a cursor nor the highlight indicator that the Matter Manipulator provides.

trey-a-12:
Hey! Apologies if this isn't allowed, I'm not really a coder or programmer, though I do enjoy the troubleshooting and research side of things. If I may, how possible is it to test said controller builds as of now? If so, how would I go about this?

If desired, I would report back any findings that could be of use, as I have both a Steam Deck and PS4 + PS5 controllers, as well as a few other miscellaneous ones like an 8BitDo Pro 2. Having just taught myself keyboard and mouse not long ago so I could finally _really_ play Starbound, I enjoyed my first playthrough, but still wish it was more feasible to explore the universe on a controller setup like the Steam Deck. Any contributions that could be done, I would gladly aide in.

Sunbrella114514:
> 嗨！如果这个问题不合适，请见谅。我其实不是程序员，但我喜欢做一些故障排除和研究方面的工作。请问现在测试这些控制器版本是否可行？如果可以，我该如何操作？
> 
> 如果需要，我可以提供任何有用的发现，因为我既有Steam Deck，也有PS4和PS5手柄，还有一些其他的手柄，比如8BitDo Pro 2。不久前我才自学了键盘和鼠标操作，这样我终于可以好_好玩_《星界边境》了。我的第一次通关体验很棒，但我仍然希望用像Steam Deck这样的手柄设备来探索宇宙会更加方便。如果有什么可以贡献的，我非常乐意帮忙。

You only need to merge this PR into your local OpenStarbound source and build the project to get a working executable.

If you're not sure how to do it, you can also give the task to an AI agent and let it handle the process for you. That's how I got it working myself.

Sunbrella114514:
I’ve been testing further and would like to report a few issues with varying reproducibility.

1. If I enter the game directly using touch input, the cursor does not appear. It only shows up if I use the right stick + trigger input before entering gameplay.

2. The Esc key does not seem to function. I can only rely on controller inputs to pause/exit. As a result, anything relying on Esc behavior (such as skip-cutscene mods or skipping respawn animations) does not work.

3. Mapping LB + RB to the “drop” action feels a bit too aggressive and is prone to accidental triggers. Additionally, when the ship is hovering near space stations or asteroid fields, the “drop” action is typically limited to mech behavior, but LB + RB does not properly trigger mech-specific drop functionality.

---

On the suggestion side (some of these may be limited by API exposure, so feel free to ignore if not feasible):

1. Star navigation planet selection feels awkward in its current state. Switching to cursor mode with right stick aiming and trigger confirmation makes interaction slower, and it conflicts with touch behavior. Ideally, in the ship navigation UI, touch input could remain functional for left-click interactions (e.g. viewing planet info), while long-press could act as right-click for initiating travel.

2. Does OSB expose any HUD-related APIs? If so, a radial HUD-based quickbar might be worth exploring. The current LB/RB cycling feels inefficient for controller use, especially when items are far apart in the hotbar. This also affects Matter Manipulator block selection in a similar way.

---

Aside from these, I don’t think there are any other major issues at the moment. I’ve mitigated aiming issues on my side using a custom “shift aimbot”-style workaround, which I’ve uploaded to the Workshop.

I’ll also attach some Xbox control layout references and HUD screenshots below for inspiration.
Screenshot_2026-06-14-12-21-41-56_149003a2d400f6adb210d7e357a3a646
Screenshot_2026-06-14-18-05-31-36_149003a2d400f6adb210d7e357a3a646
Screenshot_2026-06-14-18-04-20-27_149003a2d400f6adb210d7e357a3a646
Screenshot_2026-06-14-18-10-23-61_40deb401b9ffe8e1df2f1cc5ba480b12


Finally, I’ll include a photo of me playing Starbound on a ROG Ally. I’m genuinely grateful for enabling a better handheld experience for one of my favorite games. It means a lot to be able to revisit it in this form.
ced265c4bc3b738706392fd2979135ac_720

neonspectra:
Thanks for more feedback, all of it is helpful. Sorry, have been slow on this, work been busy. Wanted to just say haven't forgotten about this. Will look at the above when I get some time to continue working on this.

Sunbrella114514:
Thank you very much for your continued work and support!

I’ve been testing the latest progress of this branch recently, and I seem to have encountered a new issue.

When the player has the NPC Mech mod installed and deploys a mech together with crew members, logging out of the mech causes the camera to center on the crew’s mech instead of the player. This issue only occurs in controller-only input scenarios, without any keyboard or mouse input.

Building while excluding the commit “feat: extend entity userdata movement methods to support vehicles” does not appear to reproduce this problem at all.

Shadowblitz16:
Can we get a flag to compile this so it automatically enables controller support by default and disables mouse and keyboard support in options?

I would like to compile this for arm64 for my raspberry pi game console.
The issue is I don't want to have to drag my mouse and keyboard into my mothers house to switch to gamepad input so that someone can accidentally switch back to mouse and keyboard mode.

Sunbrella114514:
> 我们能否添加一个编译标志，使其默认自动启用控制器支持，并在选项中禁用鼠标和键盘支持？
> 
> 我想把它编译成 arm64 版本，用于我的树莓派游戏机。 问题是我不想每次切换到手柄输入都得把鼠标键盘搬到我妈家，以免有人不小心又切换回鼠标键盘模式。



> 我们能否添加一个编译标志，使其默认自动启用控制器支持，并在选项中禁用鼠标和键盘支持？
> 
> 我想把它编译成 arm64 版本，用于我的树莓派游戏机。 问题是我不想每次切换到手柄输入都得把鼠标键盘搬到我妈家，以免有人不小心又切换回鼠标键盘模式。

Even in the current build, controller input is enabled by default. On the main menu, you can use the right stick to emulate mouse aiming, and the left/right triggers act as click inputs, so in most cases there is no need for keyboard or mouse interaction.

It’s worth noting that a recent commit removed controller support from the main menu as well as the L3 aim mode toggle. It seems the original author is planning to refactor this system.

If you want to play on a handheld device right now, I’d recommend building the project while skipping that commit.

rafaelmomo89-create (approved):

Shadowblitz16:
> > 我们能否添加一个编译标志，使其默认自动启用控制器支持，并在选项中禁用鼠标和键盘支持？
> > 我想把它编译成 arm64 版本，用于我的树莓派游戏机。 问题是我不想每次切换到手柄输入都得把鼠标键盘搬到我妈家，以免有人不小心又切换回鼠标键盘模式。
> 
> > 我们能否添加一个编译标志，使其默认自动启用控制器支持，并在选项中禁用鼠标和键盘支持？
> > 我想把它编译成 arm64 版本，用于我的树莓派游戏机。 问题是我不想每次切换到手柄输入都得把鼠标键盘搬到我妈家，以免有人不小心又切换回鼠标键盘模式。
> 
> Even in the current build, controller input is enabled by default. On the main menu, you can use the right stick to emulate mouse aiming, and the left/right triggers act as click inputs, so in most cases there is no need for keyboard or mouse interaction.
> 
> It’s worth noting that a recent commit removed controller support from the main menu as well as the L3 aim mode toggle. It seems the original author is planning to refactor this system.
> 
> If you want to play on a handheld device right now, I’d recommend building the project while skipping that commit.

Can I get the commit that removed controller support for main menu?

Shadowblitz16:
Tested this and it needs alot more work. I guess I should have known to fork and compile a incomplete pull request.

I would suggest a few things...
- Cursor should use the block grid when using the free cursor.
- Above should probably use the dpad to move the cursor by one 
- Radial menu for accessing stuff, right now we don't have enough buttons for everything, a radial menu for quickbar actions would reduce all this to a single button.
- Add cursors for visual indicator where the mouse is.
- Ducking should be able to be bound to the left axis being pressed downwards.
- Focus based menu navigation, this is a big one but its needed for barable gameplay

Issue #456, opened 2026-02-02
Why is the server distributed as a zipped tarball?


AutumnalModding:
Surely this is inefficient? Would a gzipped tarball not be better?

If file size is a concern, `⁨7z a -t7z -m0=lzma2 -mx=7 -mmt=8 OpenStarbound.7z assets/ linux/` (for a server) produces a roughly 43MiB archive - smaller than the current zips.

It does not make sense to zip a tarball over merely zipping the files, or using a different archive format.

goodusername123:
I assume this is done as a workaround for how the default `upload-artifact` action in GitHub Actions can only output zip files and nothing else, not even raw files: https://github.com/marketplace/actions/upload-a-build-artifact#zip-archives

The solution would be to not use this to upload releases.

grbr404:
@goodusername123 your assumption is right. @AutumnalModding you can look at https://github.com/OpenStarbound/OpenStarbound/issues/245#issuecomment-2891635784 where this was already answered

DoubleHypercube:
Ah, okay. That makes sense! Thanks for the quick response.

PR #534, opened 2026-05-30
GHA-ci: optimize release compression

goodusername123:
Solves https://github.com/OpenStarbound/OpenStarbound/issues/456 by taking advantage of https://github.com/marketplace/actions/upload-a-build-artifact#upload-an-individual-file-unzipped (new feature added recently).

- ZIPs are now Deflate64 (supported natively since Windows Vista) compressed instead of Deflate.
- Windows installer is now uploaded directly instead of being inside of a ZIP.
- TARs are now Lzip (existed since 2010) compressed instead of zip (tar.lz instead of tar.zip).

The compression effort(s) for these have been set very high so if ZIP compression is considered too slow then find and replace the number in `-mx=9` with a lower value and if Lzip compression is considered too slow then find and replace the number in `-c9vf` with a lower value.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,6 +57,9 @@ jobs:
         with:
           cmakeVersion: 3.29.2
 
+      - name: setup 7-Zip action
+        uses: milliewalky/setup-7-zip@v2
+
       - name: sccache
         uses: hendrikmuhs/ccache-action@v1.2
         with:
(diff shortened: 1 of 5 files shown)


goodusername123:
Realized that I forgot to give some file size comparisons:

Linux-Clang amd64 went from around 255 MB to 210 MB
Windows Client went from around 107 MB to 97.7 MB

The Windows builds *could* be quite a bit smaller if they where say something like .7z instead of .zip however it would mean users on Windows 10 would be unable to extract the files without a third party tool such as 7-Zip, WinRAR or otherwise so I figured that such a decision would probably be better outside of this PR.

PR #535, opened 2026-05-30
Fix HDR crash

grbr404:
Fix crash: `HDR` reads `m_hdrSetting`, which was never initialized, causing a semi-consistent crash during framebuffer creation. We now set it once at init to prevent this. 

Update GLSL: Sets all GLSL versions to `150` (forgot to do this when fixing core mode).

Update checks: Updated the OpenGL version check to match our minimum supported OpenGL version


Bottinator22:
oops

PR #539, opened 2026-05-31
fix: increase LUA_IDSIZE

Mofurka:
This is so annoying when you trying to find out where is the error and stacktrace goes like this 
image
Its just folded!


Mofurka:
Ah I forgot about that my commits in main go here too

goodusername123:
@Mofurka

Would you be willing to re-submit this? on the surface the intended change here seems like a good one.

Mofurka:
@goodusername123
Oki, I'll do it today

PR #559, opened 2026-07-03
Add Shift+Enter for creating newline and display newlines as ↵

grbr404:
Adds a way for users to input a new line with `Shift+Enter` instead of just being able to paste them from an external source.

Also change newline displaying to render as `↵` instead of a space,
and adds `↵` to the hobo font.


Novaenia:
what did you use to modify the font file? there is a FontForge .sfd file in the attic folder that should be modified

grbr404:
@Novaenia i used FontForge, also the sfd file is now modified

Okay, so I just looked in the sdf files changes, and I think it looks wrong. Although I don't have much experience with changing fonts, I just added ↵ and tried to export/save somehow. Also, it's not validated, as that literally always broke when I tried to use it.

grbr404:
So re-did the font, it should now all be okay.

Issue #498, opened 2026-04-14
Super sampled AA causes game to Crash


akikoyosano:
Clicked super sampled AA on accident, game crashed, wouldnt let me open back up game cause apparently there an error and minidump was created. Had to reinstall the entire thing since I couldnt exactly turn it off since it wouldnt let me in.

SilverSokolova:
Known issue. Caused by using it with in-game shaders. Can fix without reinstall by editing starbound.config in storage folder. Might try to fix later.

PR #570, opened 2026-08-02
Feat/Lighting - Performance and Accuracy

psychosomat:
I apologize for such a lengthy PR - I originally planned to just tweak the lighting optimization a bit, but it ended up being a bit more extensive than intended. However, this patch is very valuable in terms of performance; in manual tests, the FPS gains were ~20-50%. More details on the changes are below

Before:
before
After:
after

## Restoration of ray-traced point lighting

- Xiaolin-Wu ray walk instead of Dial flood fill. Restored the original line-by-line ray tracing for point lights: the flood fill from the 3.x version “let light through” corners and doorways, resulting in a milky image without sharp shadows. Ray walk produces sharp shadows and correct light propagation through openings.
- Fixed the ray direction. The direction was only considered when beam > 0; when beam = 0, garbage was fed into the attenuation → inf → everything went dark. A single-line initialization restored torch lighting with beam = 0.
- SoA and parallel point-light phase. Light arrays were converted to an array structure (per-channel float[]), and point-light floods were distributed across a worker pool with private buffers and a merge operation. This allowed us to parallelize the most expensive phase without race conditions.
- Synchronous lighting by default. `m_asyncLighting = false` (upstream default): the async thread was returning an outdated lightmap during camera movement-causing flickering and blurring. Synchronous calculation is now efficient thanks to skip paths, diffs, and the parallel point phase.

## WorldClient Lighting Pipeline

- `lightingCalc` dispatcher. Separated into skip / incremental point-light diffs / full recalculation based on the type of change (window, tiles, ambient light, light sources, particles). Previously, almost any change triggered a full lightmap recalculation; now, frames with no changes are skipped in ~0 ms.
- Asynchronous gather with a mutex. The gathering of base light and obstacles has been moved to a preparatory thread under `m_lightMapPrepMutex` - this also resolves a race condition during sector unloading. Bulk tile reading is removed from the render thread.
- Light source culling. Light sources are filtered based on a window expanded to the maximum light radius; the number of point lights in a frame is limited by `maxPointLights` (default 128, the brightest ones are selected). Hundreds of invisible light sources are no longer counted.
- Lightmap texture is loaded only when the version changes. `lightMapVersion` controls texture re-upload - the constant re-upload every frame has been removed.

## Scrollable lightmap atlas (main performance gain)

- An atlas is used instead of recalculation when the camera moves. The light array remains an atlas the size of the “window + border”; when scrolling, data is shifted using `memmove`, and only newly exposed strips and window cells that have entered from the border are recalculated (they are cleared and repainted with the current light sources). A full recalculation occurs only when spread/tiles/env change, the window size changes, during teleportation, or after 240 incremental frames (reset of accumulated float error).
- Separate point channels. Point contributions are stored in their own array channels: otherwise, during scrolling, the spread phase would “bake” the point light from the border into the spread (max semantics), and the light would “stick” to the walls. The diff of moved sources is applied at window intersections; bands are reset to zero and filled with the current sources.
- Incremental diff of point lights. With a fixed window and changing light sources, only the remove/add operations for their contributions at window overlaps are performed. Replacing a full recalculation (~14 ms) with ~0.1–1 ms is the main source of the measured FPS gain.

## Rendering Fixes

- MSAA crash fix. Multisampling is applied only to the final framebuffer: `sampler2D` on an MSAA texture was crashing on some drivers (upstream issue #498).
- VSync option. A checkbox in the graphics menu that takes effect on the fly - added on a whim to make performance measurements easier.
- HDR in the menu. The HDR checkbox was outside the panel—moved it into the layout.

## Dependency Cleanup

- Removed dead code. Removed source/extern/tinyformat.h (1,164 lines), lib/linux/libcrypto.a (4 MB), and mimalloc from vcpkg dependencies - less code and shorter build times.
- vcpkg-overlay port of jemalloc. A port with patches for new compilers (gcc16 - jemalloc continues to build with the current vcpkg.
- `source/extern/README.md`. Documents the versions and origins of all vendor libraries - the dependency surface is auditable in one place.
- vcpkg binary cache in Windows CI. Caching package archives speeds up Windows builds in CI by about a factor of two.

## Tests and Benchmarks
- `cellular_lighting_test.cpp` - 9 tests: sharp shadows along walls and in doorways, channel separation (point does not fall into spread seeds), symmetry of add/remove diffs, equivalence of scroll path to full recalculation, scene dump to PPM for visual verification. Regression testing of all key pipeline properties.
- `cellular_lighting_bench.cpp` - micro-benchmark of lighting phases (point with 200 lights ≈ 21.7 ms, spread, skip) to monitor performance changes.

## Manual Testing

- Windows - Stable behavior
- Linux - Stable behavior
- Mac - Not tested


Bottinator22:
The HDR checkbox wasn't outside the panel for my vanilla tests, you
might've been using a modded menu. Or it was added after you already
changed the menu yourself with this PR.

emmyposs:
hello mr. openclaw bot, your MSAA/framebuffer misbehavior "fix" does not work, [as i already explained to one of you here](https://github.com/OpenStarbound/OpenStarbound/pull/561). to restate, this would be obvious to you if you simply used critical thinking skills & tested the fix with any existing shader mod.

Bottinator22:
Checks seem to have been skipped or removed as well.

PR #575, opened 2026-08-08
Installer: NTFS compress PDB files

goodusername123:
applies NTFS compression (if supported) to the debugging symbols dropped by the installer.

this reduces the final disk install size by around 331 MB


goodusername123:
closed in favor of https://github.com/OpenStarbound/OpenStarbound/pull/577 (sorry for the noise of making a second PR, I probably could have done this better!)

PR #579, opened 2026-08-12
Lua: add support for compound operators

goodusername123:
adds `+=`, `-=`, `*=`, `%=`, `/=`, `//=`, `..=`, `<<=`, `>>=`, `&=`, `|=` as compound operators to Lua's parser. `^=` is excluded on purpose due to being ambiguous.
Sourced/ported and modified/extended from:
https://github.com/paly2/ALua/#4-compound-assignment-operators-operators with light bits from http://lua-users.org/files/wiki_insecure/power_patches/5.4/plusequals-5.4.3.patch

Originally this was a extended version of https://github.com/NinjasCL-archive/lua8/#compound-assignment-operators however I realized that this patch (and it's original Lua 5.4.x counterpart(s)) seemed to behave wrong/weirdly (someone even reported as such: http://lua-users.org/wiki/LuaPowerPatches#:~:text=One%20odd%20bug%20I%20just%20found%3A) so I threw it out and rebased on [ALua](https://github.com/paly2/ALua/)'s implementation extending it from just `+=`, `-=`, and `..=` to the full list above with some of the non-broken small functional bits of the Lua 5.4.x patch ripped out to aid with extending ALua's implementation.

also here a list of other Lua implementations that contain support for compound operators:
- Pluto (superset fork of Lua 5.4/5.5): https://pluto-lang.org/docs/New%20Operators#compound-operators
- LuaJIT (heavy fork of Lua 5.1): https://luajit.org/extensions.html#lj30_bp_syntax
- Luau (heavy fork of Lua 5.1): https://luau.org/syntax/#compound-assignments
- Playdate Lua (fork of Lua 5.4): https://sdk.play.date/2.4.2/Inside%20Playdate.html#playdate-sdk-lua-enhancements
- PICO-8 (fork of Lua 5.2): https://www.lexaloffle.com/dl/docs/pico-8_manual.html#PICO_8_Shorthand


Bottinator22:
looks like this breaks a lot of stuff, including the logic for positioning the title logo and... this, for some reason.
```
[13:49:45.448] [Error] Exception while calling script init: (LuaException) Error code 2, [string "/objects/fossils/dirtyfossils/dirtyfossil.lua..."]:8: attempt to index a nil value (local 'treasure')
```

also breaks this function from TerraLib, causing RGBd to become nil
`([string "/scripts/terra_renderutil.lua"]:269: attempt to index a nil value (local 'RGBd'))`
```lua
function renderutil.toRGB(c)
    local C = c[3]*c[2]
    local X = C*(1-math.abs((c[1]/60)%2-1))
    local m = c[3]-C
    local ops = {
        {C,X,0},
        {X,C,0},
        {0,C,X},
        {0,X,C},
        {X,0,C},
        {C,0,X},
    }
    local RGBd = ops[math.floor(c[1]/60)%6+1]
    local o = {math.floor((RGBd[1]+m)*255),math.floor((RGBd[2]+m)*255),math.floor((RGBd[3]+m)*255)}
    if c[4] then
        table.insert(o,math.floor(c[4]*255))
    end
    return o
end
```

goodusername123:
> looks like this breaks a lot of stuff, including the logic for positioning the title logo and... this, for some reason.
> 
> ```
> [13:49:45.448] [Error] Exception while calling script init: (LuaException) Error code 2, [string "/objects/fossils/dirtyfossils/dirtyfossil.lua..."]:8: attempt to index a nil value (local 'treasure')
> ```
> 
> also breaks this function from TerraLib, causing RGBd to become nil `([string "/scripts/terra_renderutil.lua"]:269: attempt to index a nil value (local 'RGBd'))`
> 
> ```lua
> function renderutil.toRGB(c)
>     local C = c[3]*c[2]
>     local X = C*(1-math.abs((c[1]/60)%2-1))
>     local m = c[3]-C
>     local ops = {
>         {C,X,0},
>         {X,C,0},
>         {0,C,X},
>         {0,X,C},
>         {X,0,C},
>         {C,0,X},
>     }
>     local RGBd = ops[math.floor(c[1]/60)%6+1]
>     local o = {math.floor((RGBd[1]+m)*255),math.floor((RGBd[2]+m)*255),math.floor((RGBd[3]+m)*255)}
>     if c[4] then
>         table.insert(o,math.floor(c[4]*255))
>     end
>     return o
> end
> ```

oh that's quite worrying since this was intended to *not* break any pre-existing code. well thanks for giving this PR a go/test even if in the end it appears to be flawed/broken in some core way. In the future if I get the time I might try and track down exactly whats wrong along with doing more/actual proper testing on my end. sorry for the noise!

goodusername123:
@Bottinator22

After checking the code/diff more closely I have come to realize that I mistakenly broke a good chunk of the basic non-compound versions of operators so that they return the result of `+` rather then the intended result (WHOOPS). guess that's what I get for pushing while low on sleep.
I have pushed a commit to the branch this PR is based on to fix this: https://github.com/goodusername123/OpenStarbound/commit/c6769565bb58989d35e6c3ec2f3b2cbe2c17dbc2 unsure of how Github handles commits pushed to a branch after a PR getting marked as merged but you should be able to cherry pick the commit itself (or just manually make a new commit with the same changes since it's so small) or alternatively I could submit a new PR if desired.

I apologize for this glaring oversight on my part, sorry for the hassle.

PR #580, opened 2026-08-13
fix rendering issues on small worlds.

grbr404:
So, after a bit more looking through the code, I have found the issues causing the rendering bug mentioned in my last PR and fixed all cases of it. In most cases, it was just that `diff` was used where `pdiff` should have been used. The lighting calc was also changed to get its position now through `pdiff`, as it had the same issue, even though it didn't use `diff`.

Also, whilst I'm at it, could you take a look at #560? I have changed its log messages to be better. If there still is an issue with it, I would appreciate a comment.

### DIFF
--- a/source/game/StarTileDrawer.cpp
+++ b/source/game/StarTileDrawer.cpp
@@ -176,7 +176,7 @@ MutexLocker TileDrawer::lockRenderData() {
 }
 
 RenderTile const& TileDrawer::getRenderTile(WorldRenderData const& renderData, Vec2I const& worldPos) {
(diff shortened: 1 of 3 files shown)


Bottinator22:
Does this fix the segfault crash that occurs when the screen is wider than the world?

grbr404:
@Bottinator22  just tried and no it dosnt fix the segfault only the rendering issues as i described. the segfault is probably related to the wrapping functions. https://github.com/OpenStarbound/OpenStarbound/blob/0287e9c5a9ca56b0e64ec580ea40d7442f3cba54/source/base/StarWorldGeometry.hpp#L71-L72

if you want to look at what this fixes i explained the issue in #578

PR #583, opened 2026-08-20
Client subworlds

Bottinator22:
Allows the client to create subworlds. 

Subworlds can coexist with the main world, the client can warp around and the subworlds will remain. Subworlds disconnect when the client disconnects.
Subworlds automatically unload the world after 10 seconds of no entities or messages.
Subworld threads are automatically stopped after 10 seconds of not having any world loaded.

Subworlds initially start with no entities. It is up to worldClient script contexts (also added in this PR) to do anything with the worlds or respond to messages. The main client world is also allowed contexts and message handling.
Subworlds are functionally similar to server worlds.

`ClientPresenceMaster` entities are required to load anything on subworlds, as they have no players and no windows.
Subworlds use the second half of the ConnectionId range, with their world-specific ConnectionId being the second half of the players'. This lowers the player limit to 16k (which is still a lot for one Starbound server to handle) but allows retaining the property of every entity within a given 65k client range being considered same master.

Subworlds can be configured with the following server config values:
`disallowClientSubWorlds` (defaults false) disables the entire system.
`allowClientSubWorldsAllWorlds` (defaults true) allows creating subworlds on any world.
`allowClientSubWorldsCurrentWorld` (defaults true) only allows creating subworlds on the client's current world.
If not disallowed, subworlds can always be created on the client's own ship/custom worlds.

Additional features, useful for subworlds:
- Clients can be notified on serverside world create, invoking a callback on the client's UniverseClient script contexts. This always happens for the clients' own custom worlds and shipworld, and can optionally happen for all worlds, optionally for admins only. Configured with server config values `notifyClientsOnWorldCreate` (defaults false) and `notifyAdminsOnWorldCreate` (defaults false)
- Some additional Lua bindings

Documentation for relevant Lua bindings pending.

### DIFF
--- a/assets/opensb/client.config.patch
+++ b/assets/opensb/client.config.patch
@@ -11,5 +11,12 @@
     "letterbox" : false
   },
   "postProcessLayers": [],
-  "postProcessGroups": {}
+  "postProcessGroups": {},
+  
+  "worldScriptContexts":{},
+  "mainWorldScriptContexts":{},
+  "subWorldScriptContexts":{},
+  
+  "idleSubWorldExpireTime":10.0,
+  "subWorldUpdateMeasureWindow" : 0.5
 }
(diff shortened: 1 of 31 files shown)


DarianLStephens:
Huh, what sort of things could this be used for?

PR #592, opened 2026-08-26
Allow using vendored imgui

fpiesche:
Currently, imgui is the only dependency for OpenStarbound that *needs* to be pulled in by vcpkg - imgui itself does not have a module for finding the library via cmake, as its intended distribution vector is as a vendored library. This also means that Linux builds, which can use system-provided libraries for all other dependencies, currently still require imgui to be installed via vcpkg.

This prevents OpenStarbound being built in restrictive environments (eg. building the server in an Alpine Docker container where installing vcpkg and using it for dependencies massively inflates build times and image sizes, or in a Flatpak distribution for the client on Linux where network connectivity is disabled during builds as part of the sandboxing).

With this change, a vendored copy of imgui in `source/extern/imgui/` can be used instead by setting `-DSTAR_USE_VENDORED_IMGUI=ON`, removing the need for vcpkg and pulling things off the internet during build time.


Bottinator22:
This fails to build in the workflow runs.

fpiesche:
> This fails to build in the workflow runs.

For some reason the workflows on my repo have just been stalled waiting to run and refusing to let me cancel or re-run them so ta! Will investigate.

Edit: ah - I've been running my local builds with gcc rather than clang, that might be the issue here. Will give that a go and fix it up...

fpiesche:
The problem turned out to be a lot simpler than I thought; the setup for `STAR_USE_JEMALLOC` would just overwrite the imgui entry in the `STAR_EXTERN_LIBS` list I set earlier.

Issue #597, opened 2026-08-29
Please support opengl es for raspberry pi 5


Shadowblitz16:
I can't get open starbound to run because it fails with a gl config error.

```
#!/bin/bash
export DISPLAY=:0.0
export SDL_VIDEODRIVER=kmsdrmexport SDL_VIDEO_GL_DRIVER=/usr/lib/aarch64-linux-gnu/libGL.so.1export LIBGL_ALWAYS_SOFTWARE=0
sh /userdata/system/add-ons/starbound/linux/run-client.sh
```
|
v
```
[08:32:49.055] [Info] Root: Preparing...
[08:32:49.056] [Info] Root: Done preparing Root.
[08:32:49.056] [Info] OpenStarbound Client v0.1.14.1 for v1.4.4 (linux arm64) Source ID: fa6df17d7096ecadd49ba9db2323f4516822d7a0 Protocol: 747
[08:32:49.056] [Info] Application: Initializing SDL Video
[08:32:50.510] [Info] Root: Shutting down Root
[08:32:50.512] [Error] Fatal Exception caught: (ApplicationException) Couldn't initialize SDL Video: kmsdrmexport not available
Stack trace (most recent call first):
#0 (inlined)          in captureBacktrace at StarException_unix.cpp:20:48
#1 0x0000000000cbe5ab in Star::StarException::StarException(char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool) at StarException_unix.cpp:98:91
#2 0x000000000184d137 in Star::ApplicationException::ApplicationException(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, bool) at StarApplication.hpp:17:1
#3 0x0000000001854877 in Star::SdlPlatform::SdlPlatform(std::unique_ptr<Star::Application, std::default_delete<Star::Application> >, Star::StringList) at StarMainApplication_sdl.cpp:487:7
#4 0x000000000184b757 in Star::runMainApplication(std::unique_ptr<Star::Application, std::default_delete<Star::Application> >, Star::StringList) at StarMainApplication_sdl.cpp:1401:74
#5 0x0000000000b29743 in main at StarClientApplication.cpp:1917:1
#6 0x0000007fb79eb4ff at libc.so.6
#7 0x0000007fb79eb5d7 in __libc_start_main at libc.so.6
#8 0x0000000000b2c2af in _start at starbound

Caught at:
[...]

PR #598, opened 2026-08-29
add 3.1 to supported OpenGL context

grbr404:
So, as in the title, this adds GL 3.1 to the supported context. As it is the first version with GLSL 140.

The GLSL version was updated to 150 in the past due to a misunderstanding of why 140 was used, which is why I reverted this mistake I made with this PR.

This might help with #597 because it seems like it at a maximum supports 3.1

Also, I removed the OpenGL version check in the OpenGL renderer. The only reason this code was ever there is because the original code didn't set a context, so the system chose one, and therefore the renderer needed to check what it got.

### DIFF
--- a/assets/opensb/rendering/effects/basic.vert
+++ b/assets/opensb/rendering/effects/basic.vert
@@ -1,4 +1,4 @@
-#version 150
+#version 140
 
 in vec2 vertexPosition;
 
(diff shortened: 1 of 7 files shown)


grbr404:
For context, this PR mainly just tries to lower the minimum requirement and might help with the issue. However, I will not try to find a way to get OpenGL ES to work as I don't have a system that supports it.

PR #619, opened 2026-09-14
Add chat color helper utility (colorText/colorTexts)

Reuno92:
# Summary

Adds a small utility script (`assets/opensb/scripts/opensb/util/chatcolor.lua`) with two helpers for Starbound's inline chat color escape codes (`^colorname;text^reset;`):

- `colorText(text, color)` — wraps a single string in a color.
- `colorTexts(texts, colors)` — wraps a list of strings, each in its own  color from a parallel `colors` list, and joins the result. Falls back to a default color if `colors` has fewer entries than `texts`.

# Motivation

The color escape code syntax isn't obvious for mod authors (easy to confuse with `npc.say()`'s unrelated `color = {r,g,b}` bubble option), and there's currently no shared helper for it alongside the existing `util/modules.lua`. This came out of writing chat feedback messages for a Starbound mod and wanting a reusable, correct helper instead of hand-building escape strings each time.

# Example

Simple example:
```lua
chat.addMessage(colorText("Saved!", "green"), { mode = "Local" })
chat.addMessage(colorTexts({"H", "e", "l", "l", "o"}, {"red", "orange", "yellow", "green", "blue"}), { mode = "Local" })
```

Real life example:
```lua
chat.addMessage(colorTexts({ "[Mod] ", "alert ", "An error has been encountered" }, {"yellow", "red" }), { mode = "Local" })
```

[Rendered output](https://claude.ai/code/artifact/0c5160ea-b04f-4722-b9cb-e960e12855a4) for each example above


Lonaasan:
you code makes bloated colorstrings (always adding ^reset; after each. + there are more than just colors that text can use like ^font, ^shadow or even directives). in general, this might *look* nice for a human reader (and even that not really if you have long texts with many colors. imagine trying to find out which part of the text table is linked to the color one), but is not really anything *needed*.

Bottinator22:
This is out of scope for oSB. If a mod needs this script, the mod can very easily include it itself or in a dependency.

Same reason I'm not adding all of TerraLib to oSB.

